PR #2, opened 2015-08-23
flavor downloader

sabanas:


lib/DownloadUtils.js
```diff
@@ -0,0 +1,58 @@
+/**
+ * Created by AsherS on 8/20/15.
+ */
+
+var fs = require('fs');
```
eladBenedict: 1. Please move utils to a 'utils' folder
2. I'm pretty sure qHttp.request.read does exactly what you're looking for here [https://www.npmjs.com/package/q-http]

lib/DownloadUtils.js
```diff
@@ -0,0 +1,58 @@
...
+var request = require('request');
+var Q = require('q');
+
+var DownloadUtils = (function(){
+
+    function DownloadUtils(){
+
+    }
+
+    DownloadUtils.downloadFile = function (url,dest) {
```
eladBenedict: I think destPath is preferable so it's clear this is an entire path

lib/DownloadUtils.js
```diff
@@ -0,0 +1,58 @@
...
+                    req.pipe(file);
+                })
+
+                .on('error', function (err) {
+                    console.log(err);   //TODO error msg
+                    deferred.reject(err);
+                })
+
+                .on('complete', function () {
+                    deferred.resolve(); //TODO resolve value?
```
eladBenedict: This marks the operation as successful, although onComplete only marks the fetching as successful, not the entire downloading process

lib/DownloadUtils.js
```diff
@@ -0,0 +1,58 @@
...
+
+    DownloadUtils.downloadFile = function (url,dest) {
+
+        var deferred = Q.defer();
+
+        try {
+            var req = request.get(url);
+            req
+                .on('response', function (response) {
+                    console.log("url : " + url);
```
eladBenedict: All console.log should be migrated to winston.log

lib/DownloadUtils.js
```diff
@@ -0,0 +1,58 @@
...
+ * Created by AsherS on 8/20/15.
+ */
+
+var fs = require('fs');
+var request = require('request');
+var Q = require('q');
+
+var DownloadUtils = (function(){
+
+    function DownloadUtils(){
```
eladBenedict: You have a function which when you call returns a function whose sole purpose is to host the "downloadFile" function - this is unnecessarily complex - why not just return an object with a downloadFile function (assuming you decide not to use q-http)?

lib/FileUtils
```diff
@@ -0,0 +1,33 @@
+/**
+ * Created by AsherS on 8/20/15.
+ */
+
+var fs = require('fs');
+var Q = require('q');
+
+var FileUtils = (function(){
+
+    function FileUtils(){
```
eladBenedict: same here

lib/FileUtils
```diff
@@ -0,0 +1,33 @@
...
+
+    }
+
+    FileUtils.listDirFiles = function(path) {
```
eladBenedict: I'm pretty sure q-io.listTree [https://github.com/kriskowal/q-io] does exactly this

lib/HLSDownloader.js
```diff
@@ -0,0 +1,113 @@
+/**
+ * Created by AsherS on 8/20/15.
```
eladBenedict: Duplicate comment

lib/HLSDownloader.js
```diff
@@ -0,0 +1,113 @@
...
+var hlsUtils = require('./HLSUtils');
+
+var HLSDownloader = (function () {
+
+    function HLSDownloader(m3u8Url, destPath) {
+        this.streamUrl = m3u8Url;
+        this.destPath = destPath;
+    }
+
+    var downloadIndefinitely = function (url, dest) {
```
eladBenedict: Downloading start/stop should be triggered by the entry downloader

lib/HLSDownloader.js
```diff
@@ -0,0 +1,113 @@
...
+        downloadPlaylist(url, dest)
+            .then(function(targetDuration){
+                console.log('results: ' + targetDuration);
+            })
+            .catch(function(err){
+                console.log('error ' + err);
+            })
+            .done();
```
eladBenedict: .done() has no meaning if you .catch beforehand.
Also, the polling interval should be configurable (as you mentioned)

lib/DownloadUtils.js
```diff
@@ -0,0 +1,58 @@
...
+
+    }
+
+    DownloadUtils.downloadFile = function (url,dest) {
+
+        var deferred = Q.defer();
+
+        try {
+            var req = request.get(url);
+            req
```
nirbe: What's request max connection value? If it uses http internally, than the Node.js http client automatically uses socket pooling: by default, this limits you to 5 sockets per host which is quite low for downloading ts files from the same Wowza host.  Consider enlarging it. See http://nodejs.org/docs/v0.6.3/api/http.html#agent.maxSockets

lib/HLSDownloader.js
```diff
@@ -0,0 +1,113 @@
...
+            console.log('-------------------');
+            downloadIndefinitely(url, dest);
+        }, 10000);  //todo, configuration?
+    };
+
+    var downloadPlaylist = function (m3u8Url, dest) {
+
+        var parsedUrl = url.parse(m3u8Url);
+        var baseUrl = parsedUrl.href.substring(0, parsedUrl.href.lastIndexOf('/'));
+        var playlistName = path.basename(parsedUrl.pathname);
```
eladBenedict: playlistName will soon be taken from a central object (a.k.a PersistenceFormat)

lib/HLSDownloader.js
```diff
@@ -0,0 +1,113 @@
...
+            })
+            .then(function(m3u) {
+                targetDuration = m3u.get("targetDuration");
+                downloadedM3u8 = m3u;
+                return fileUtils.listDirFiles(dest);
+            })
+            .then(function(files) {
+                //get all files that are in playlist but not in dir
+                //TODO - null pointer on playlistItem
+                return _.chain(downloadedM3u8.items.PlaylistItem).map(function(i){
```
eladBenedict: URIs are sometimes absolute and sometimes relative - you should address it

lib/HLSDownloader.js
```diff
@@ -0,0 +1,113 @@
...
+
+                    //return the promise of the downloaded chunk
+                    return downloadUtils.downloadFile(chunkUrl, path.join(dest,chunk));
+                }).value();
+            })
+            .then(function resolveAllDownloads(promises) {
+                return Q.allSettled(promises);
+            })
+            .then(function(results) {
+                return _.filter(results, function(p) {
```
eladBenedict: 1. Need to update M3U8 with new chunks
2. We need to see how we handle a case where we had an error downloading a chunk - * The M3U8 should always have chunks in the correct order! *

lib/HLSDownloader.js
```diff
@@ -0,0 +1,113 @@
...
+                    return (p.state !== 'fulfilled');
+                })
+            })
+            .then(function checkRejectedPromises(filteredResults) {
+                console.log("Filtered results: " + filteredResults.length);
+                return [filteredResults.length === 0, targetDuration];
```
eladBenedict: I suggest we ignore targetDuration altogether - let's just have a configurable polling interval - it's simpler and IMO just as good

eladBenedict: Why not reject with the failing files? (return Q.reject(filteredResults))

package.json
```diff
@@ -1,22 +1,24 @@
...
+    "matchdep": "~0.3.0",
+    "grunt-mocha-test": "0.9.3",
+    "jshint-junit-reporter": "0.0.6",
+    "mocha-html-reporter": "0.0.1",
+    "sinon": "1.9.0",
+    "sinon-chai": "2.5.0",
+    "m3u8": "0.0.6",
+    "q-io": "1.13.1",
+    "mocha": "2.2.5",
+    "q": "1.4.1",
```
eladBenedict: q, q-io, m3u8 and underscore are also "real" dependencies

lib/HLSDownloader.js
sabanas: here i assume that i have the complete url to the m3u8 file, hence, i already have the playlist's name

lib/HLSDownloader.js
sabanas: i should address it here only if the chunks in the m3u8 are urls and not just chunklist names. i haven't seen any such case so far. what do u think?

PR #3, opened 2015-08-25
Genesis flavor downloader

sabanas:


lib/flavor-downloader.js
```diff
@@ -0,0 +1,144 @@
...
+            .catch(function (err) {
+                logger.error('Error occurred in download iteration: ' + err);
+            });
+
+        setTimeout(function () {
+            if (that.shouldRun) {
+                downloadIndefinitely.call(that);
+            }
+            else {
+                logger.info('downloader has shut down');
```
eladBenedict: Specify the entry and flavor (in all logging)

lib/flavor-downloader.js
```diff
@@ -0,0 +1,144 @@
...
+                        return !_.contains(files, item.get('uri'));
+                    }).value();
+
+                //generate updated m3u8
+                m3u8generator.update(chunksToDownload);
+
+                //download chunks
+                return _.map(chunksToDownload, function (chunk) {
+                    var chunkName = chunk.get('uri');
+                    var chunkUrl = url.resolve(baseUrl + '/', chunkName);
```
eladBenedict: Does this also work for absolute chunk URIs?

lib/flavor-downloader.js
```diff
@@ -0,0 +1,144 @@
...
+var FlavorDownloader = (function () {
+
+    function HLSDownloader(m3u8Url, destPath) {
+        this.streamUrl = m3u8Url;
+        this.destFolderPath = destPath;
+        this.shouldRun = true;
+        this.pollingInterval = config.get("pollingInterval");
+        this.tmpM3u8Folder = path.join(destPath, 'tmpM3u8');
+        this.playlistName = path.basename(url.parse(m3u8Url).pathname);
+        this.m3u8generator = chunklistM3u8Generator(this.destFolderPath, this.playlistName, config.get("dvrWindow"));
```
eladBenedict: playlist name (and flavor destination folder) should be taken from PersistenceFormat

lib/flavor-downloader.js
```diff
@@ -0,0 +1,144 @@
...
+        var downloadedM3u8 = undefined;
+
+        return httpUtils.downloadFile(m3u8Url, playlistDestPath)
+
+            .then(function finishedDownload() {
+                return m3u8Handler.parseM3U8(playlistDestPath);
+            })
+
+            .then(function listFilesInDest(m3u) {
+                logger.debug("Received M3U8: " + m3u);
```
eladBenedict: m3u.toString()

lib/flavor-downloader.js
```diff
@@ -0,0 +1,144 @@
...
+                downloadedM3u8 = m3u;
+                return qfs.list(filesDestFolder);
+            })
+
+            .then(function downloadMissingChunks(files) {
+                //get all files that are in playlist but not in dir
+                var chunksToDownload =  _.chain(downloadedM3u8.items.PlaylistItem)
+
+                    .filter(function(item) {
+                        return !_.contains(files, item.get('uri'));
```
eladBenedict: This will definitely not work for absolute URI

lib/flavor-downloader.js
```diff
@@ -0,0 +1,144 @@
...
+                    }).value();
+
+                //generate updated m3u8
+                m3u8generator.update(chunksToDownload);
```
eladBenedict: update returns a promise - don't ignore it

lib/flavor-downloader.js
```diff
@@ -0,0 +1,144 @@
...
+                    var chunkName = chunk.get('uri');
+                    var chunkUrl = url.resolve(baseUrl + '/', chunkName);
+
+                    //return the promise of the downloaded chunk
+                    return httpUtils.downloadFile(chunkUrl, path.join(filesDestFolder, chunkName));
+                });
+            })
+
+            .then(function resolveAllDownloads(promises) {
+                return Q.allSettled(promises);
```
eladBenedict: No need for this 'then' - in the previous one write:
var promises = _.map(....)
return Q.allSettled(promises);

lib/flavor-downloader.js
```diff
@@ -0,0 +1,144 @@
...
+
+            .then(function filterResolvedPromises(results) {
+                return _.filter(results, function(p) {
+                    return (p.state !== 'fulfilled');
+                })
+            })
+
+            .then(function checkRejectedPromises(filteredResults) {
+                if (filteredResults.length > 0) {
+                    return Q.reject(filteredResults)
```
eladBenedict: always reject with an Error instance. On that instance you can place the list of erronous downloads

lib/flavor-downloader.js
```diff
@@ -0,0 +1,144 @@
...
+    HLSDownloader.prototype.stop = function () {
+        this.shouldRun = false;
+    };
+
+    return HLSDownloader;
+})();
+
+
+
+var downloader = new FlavorDownloader("http://kalsegsec-a.akamaihd.net/dc-0/m/pa-live-publish3/kLive/smil:1_oorxcge2_publish.smil/chunklist_b475136.m3u8", "/Users/AsherS/Downloads/DVR").start();
```
eladBenedict: Remove this line

lib/logger/logger.js
```diff
@@ -0,0 +1,33 @@
...
+            handleExceptions: true,
+            json: false,
+            colorize: true
+        })
+    ],
+    exitOnError: false
+});
+
+module.exports = logger;
+module.exports.stream = {
```
eladBenedict: What is this for?

lib/logger/logger.js
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+var winston = require('winston');
+winston.emitErrs = true;
+
+var logger = new winston.Logger({
+    transports: [
+        new (winston.transports.File)({
+            name: 'info-file',
+            filename: 'logs/filelog-info.log',
```
eladBenedict: please take this from configuration as well (level and dest folder)

lib/utils/HTTPUtils.js
```diff
@@ -0,0 +1,45 @@
...
+ * Created by AsherS on 8/24/15.
+ */
+
+var qfs = require("q-io/fs");
+var networkClient = require('../NetworkClientFactory').getNetworkClient();
+var config = require('../Configuration');
+
+var HTTPUtils = (function(){
+
+    var obj = {};
```
eladBenedict: you can just write 
return {
downloadFile : function(...) {...}
}
but it's up to you :)

lib/utils/HTTPUtils.js
```diff
@@ -0,0 +1,45 @@
...
+
+    obj.downloadFile = function (url,pathDestination, timeout) {
```
eladBenedict: space after url,

lib/utils/HTTPUtils.js
```diff
@@ -0,0 +1,45 @@
...
+            request.timeout = timeout;
+        }
+        else if (config.get("requestTimeout")) {
+            request.timeout = config.get("requestTimeout");
+        }
+
+        return networkClient
+            .read(request)
+            .then(
+            function(res) {
```
eladBenedict: You can delete this then - it's meaningless (it does nothing)

lib/flavor-downloader.js
sabanas: flavour downloader gets all params from entry downloader which will be responsible for getting the correct url/path via persistanceFormat

lib/flavor-downloader.js
sabanas: except logging (success/fail), i have nothing practical to do with that promise.
would you like to log the status or do something else with it?

sabanas: .

lib/flavor-downloader.js
```diff
@@ -0,0 +1,140 @@
...
+            .then(function downloadMissingChunks(files) {
+                //get all files that are in playlist but not in dir
+                var chunksToDownload =  _.chain(downloadedM3u8.items.PlaylistItem)
+
+                    .filter(function(item) {
+                        return !_.contains(files, item.get('uri'));
+                    }).value();
+
+                //generate updated m3u8
+                that.m3u8generator.update(chunksToDownload);
```
sabanas: except logging (success/fail), i have nothing practical to do with that promise.
would you like to log the status or do something else with it?

lib/flavor-downloader.js
eladBenedict: playlistName is calculated based on the *live* m3u8Url. How will it get the correct manifest name originating from PersistenceFormat? What am I missing?

lib/flavor-downloader.js
eladBenedict: Each iteration is comprised of 2 steps (which can be executed in parallel once the missing chunk list is calculated) - 
1. download missing chunks
2. update missing chunks in m3u8
if any of these fails your overall function (downloadPlaylist) should succeed or fail accordingly

lib/flavor-downloader.js
```diff
@@ -0,0 +1,140 @@
...
+            if (that.shouldRun) {
+                downloadIndefinitely.call(that);
+            }
+            else {
+                logger.info(formatLogMsg(that,'downloader has shut down'));
+            }
+        }, that.pollingInterval);
+    };
+
+    var downloadPlaylist = function (that) {
```
eladBenedict: you can remove the "that" parameter and call downloadPlaylist with downloadPlaylist.call(this) the same way you do with downloadIndefinitely

tests/http-utils-spec.js
```diff
@@ -0,0 +1,100 @@
+/**
```
eladBenedict: Where are the flavourDownloader tests? Did you forget to check them in?

PR #10, opened 2015-08-27
Generate urls properly

eladBenedict:


lib/MasterManifestGenerator.js
```diff
@@ -8,45 +8,61 @@ var path = require('path');
...
 var _ = require('underscore');
 var networkClient = require('./NetworkClientFactory').getNetworkClient();
+var url = require('url');
+var config = require('./Configuration');
 
 module.exports = function(entryId, hostname, port, applicationName) {
 
+    /* jshint shadow:true */
     var persistenceFormat = require('./PersistenceFormat');
+    var hostname = hostname || 'localhost';
```
tan-tan-kanarek: The var is redundant for hostname, port and applicationName.

PR #12, opened 2015-08-30
Add worker

eladBenedict:


lib/config/config.json
```diff
@@ -1,8 +1,18 @@
 {
   "mockNetwork" : false,
+  "mockBackend" : true,
+  "rootFolderPath" : "c:\\dvr",
```
tan-tan-kanarek: Please use linux style path or relative path if possible.

PR #18, opened 2015-09-06
Add session manager

eladBenedict:


lib/SessionManager.js
```diff
@@ -0,0 +1,49 @@
...
+var config = require('./Configuration');
+var Q = require('q');
+var qio = require('q-io/fs');
+var _ = require('underscore');
+
+module.exports = (function(){
+
+    var isNewSession = function isNewSession(entryId) {
+        return Q.fcall(function () {
+            var entryFolder = persistenceFormat.getEntryFullPath(entryId);
```
sabanas: this is a sync call that only concatenate strings. would you still like to wrap it as a promise?

lib/SessionManager.js
```diff
@@ -0,0 +1,49 @@
...
+            return entryFolder;
+        }).then(function (entryFolder) {
+            var getManifestFiles = Q.denodeify(glob);
+            return getManifestFiles("**/*.m3u8", {cwd: entryFolder, realpath: true});
```
sabanas: why list all files? you already know the m3u8 file name  - persistanceFormat. getManifestName()

lib/SessionManager.js
eladBenedict: Yes - I want to ensure this function always returns a promise (and never throws an exception)

lib/SessionManager.js
eladBenedict: You're right, however in this (session manager's) context I don't know what flavors are available for the entry.
I could have passed those as part of the session manager's API, however I think the current API (accepting only an entry id) is more sound (For example, when we'll change the logic to infer a new session based on backend entry data it won't have to change)

PR #24, opened 2015-09-20
Adding WS configuration + moving common code to common folder

eladBenedict:


common/config/config.json
```diff
@@ -12,16 +12,20 @@
...
+    "port" : 8081,
+    "logLevel" : "debug",
+    "logFileName" : "c:\\dvr\\ws.log"
   },
   "backendClient" : {
     "adminSecret" :"admin secret",
-    "partnerId" : "partner Id (integer)",
-    "serviceUrl" : "back end service url",
-    "ksSessionRefreshIntervalMinutes" : 60
+    "partnerId" : 12345,
```
tan-tan-kanarek: IMO, the place-holder was better.
I would also remove the hostname, and the windows paths from this **public** code.

PR #33, opened 2015-09-27
Change default log path to /var/log

eladBenedict:


common/config/config.json.template
```diff
@@ -11,7 +11,7 @@
   "liveDvrWindow" : 100,
   "logLevel" : "debug",
   "logToConsole" : true,
-  "logFileName" : "../logs/dvr.log",
```
tan-tan-kanarek: This path used only on our SaaS servers, on any other environment we use /opt/kaltura/log/kaltura...
For the logs dir we use common token, so I suggest to replace with @LOG_DIR@/kaltura_dvr_worker.log and @LOG_DIR@/kaltura_dvr_ws.log.
See example: https://github.com/kaltura/server/blob/Jupiter-10.21.0/configurations/logger.template.ini.

PR #47, opened 2015-11-08
Add backend client unit tests

ronyad:
add more unit tests to backend client.

### DIFF
--- a/lib/BackendClient.js
+++ b/lib/BackendClient.js
@@ -34,7 +34,6 @@ var BackendClient = (function(){
     var BackendClient = {};
 
     var createSession = function () {
-
         var deferred = Q.defer();
         client = new kalturaClient.KalturaClient(clientConfig);
         client.session.start(function(ks, err){
@@ -100,6 +99,12 @@ var BackendClient = (function(){
             return getLiveEntries();
         }
     };
+    BackendClient.SetLastSessionTime=function(TimeTOSet){
+        lastSessionTime=TimeTOSet;
+    };
+    BackendClient.GetlastSessionTime=function(){
+        return lastSessionTime;
+    }
     return BackendClient;
 })();
 
--- a/lib/BackendClient.js
+++ b/lib/BackendClient.js
@@ -99,10 +99,10 @@ var BackendClient = (function(){
             return getLiveEntries();
         }
     };
-    BackendClient.SetLastSessionTime=function(TimeTOSet){
-        lastSessionTime=TimeTOSet;
+    BackendClient._setLastSessionTime=function(timeToSet){
+        lastSessionTime=timeToSet;
     };
-    BackendClient.GetlastSessionTime=function(){
+    BackendClient._getlastSessionTime=function(){
         return lastSessionTime;
     }
     return BackendClient;
--- a/tests/unit/BackendClient-spec.js
+++ b/tests/unit/BackendClient-spec.js
@@ -1,0 +1,134 @@
+/**
+ * Created by Ron.Yadgar on 11/3/2015.
+ */
+var proxyquire = require('proxyquire');
+var chai = require('chai');
+var expect = chai.expect;
+var sinon = require('sinon');
+var config = require('../../common/Configuration');
+describe('Backend client factory spec', function() {
+    var text;
+    /*
+     var loggerMock = {
+     info: sinon.stub(),
+     error: sinon.stub(),
+     debug: sinon.stub(),
+     warn: sinon.stub()
+     };*/
+    var loggerMock = {
+        info: function(str){  console.log("info: "+str)},
+        error:   function(){ console.log("error")},
+        debug:  function(){  console.log("debug")},
+        warn:   function(){ console.log("warn")}
+    };
+
+
+
+    var createBackendClient = function(mocks) {
+        var BackendClientCtor = proxyquire('../../lib/backendClient', mocks);
+        return BackendClientCtor;
+    };
+    //
+    //it("Check that createSession function is called after session time is expired", function () {
+    //    KalturaClientMock = {
+    //        KalturaClient: function(){
+    //            this.session={
+    //                'start':functionIndicator
+    //            };
+    //        }
+    //    };
+    //    functionIndicator=function(){
+    //        return  sinon.spy();
+    //    }
+    //    var mocks = {
+    //        './kaltura-client-lib/KalturaClient' : KalturaClientMock,
+    //        './logger/logger': sinon.stub().returns(loggerMock)
+    //    };
+    //    clock = sinon.useFakeTimers(9999999999);
+    //    var BackendClient=createBackendClient(mocks);
+    //    BackendClient.SetLastSessionTime(0);
+    //    BackendClient.getLiveEntriesForMediaServer();
+    //    expect(loggerMock.debug).to.have.been.calledOnce;
+    //});
+    it ("Session expired, check for creating new session",function(done){
+      KalturaClientMock = {
+            KalturaClient : sinon.stub().returns({
+                session : {
+                    start : sinon.stub().callsArgWith(0,'1')
+                },
+                setSessionId: sinon.stub()
+            })
+        };
+        var mocks = {
+            './logger/logger': sinon.stub().returns(loggerMock),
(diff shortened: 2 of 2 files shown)


kaltura-hooks:
Hi @ronyad,
Thank you for contributing this pull request!
Please sign the Kaltura CLA so we can review and merge your contribution.
Learn more at http://bit.ly/KalturaContrib

kaltura-hooks:
Hi @ronyad,
Thank you for contributing this pull request!
Please sign the Kaltura CLA so we can review and merge your contribution.
Learn more at http://bit.ly/KalturaContrib

lib/BackendClient.js
```diff
@@ -100,6 +99,12 @@ var BackendClient = (function(){
             return getLiveEntries();
         }
     };
+    BackendClient.SetLastSessionTime=function(TimeTOSet){
```
eladBenedict: 1. use camelCased function and argument names - "setLastSessionTime", "timeToSet", etc.
2. since these are basically private, use _ as the name prefix (e.g. _setLastSessionTime)
3. There's no real gain from the getter\setter method - I'd go with the variable itself or define a property (https://msdn.microsoft.com/en-us/library/dd548687(v=vs.94).aspx)

tests/unit/BackendClient-spec.js
```diff
@@ -0,0 +1,183 @@
+/**
+ * Created by Ron.Yadgar on 11/3/2015.
+ */
+var proxyquire = require('proxyquire');
+var chai = require('chai');
+var expect = chai.expect;
+var sinon = require('sinon');
+var config = require('../../common/Configuration');
+var assert=require('assert')
+describe('Backend client facdefgstory spec', function() {
```
eladBenedict: 1. Add a blank line after the requires for readability
2. describe name is gibberish

tests/unit/BackendClient-spec.js
```diff
@@ -0,0 +1,183 @@
...
+                done();
+            },
+            function(err)
+            {
+                console.log('Promise failed');
+                done(err);
+            });
+    });
+    it ("Promise should be failed, after error occurred, in the create new session",function(done){
+        KalturaClientMock = {
```
eladBenedict: Move this object along with the mocks object to the createBackendClient function since they all pretty much use the same code. In your tests where you wish to override something, do so via a custom function - look at createManifestGenerator(customizeMocks, ... ) in ChunklistManifestGenerator-spec.js as an example.
Then, you can just override the relevant method (e.g. mocks['./someModule'].someStub.returns(newValue);

tests/unit/BackendClient-spec.js
```diff
@@ -0,0 +1,183 @@
...
+                expect(BackendClient.GetlastSessionTime()).to.eql(fakeTime);
+                done();
+            },
+            function(err)
+            {
+                console.log('Promise failed');
+                done(err);
+            });
+    });
+    it ("Promise should be failed, after error occurred, in the create new session",function(done){
```
eladBenedict: Rephrase - Should return a rejected promise when create session fails

tests/unit/BackendClient-spec.js
```diff
@@ -0,0 +1,183 @@
...
+                    start : sinon.stub().callsArgWith(0,null,'1')
+                },
+                setSessionId: sinon.stub()
+            })
+        };
+        var mocks = {
+            './logger/logger': sinon.stub().returns(loggerMock),
+            './kaltura-client-lib/KalturaClient' : KalturaClientMock
+        };
+        var sessionDurationTest=config.get('backendClient').ksSessionRefreshIntervalMinutes* 60 * 1000;
```
eladBenedict: config is another dependency - stub it as well

tests/unit/BackendClient-spec.js
```diff
@@ -0,0 +1,183 @@
...
+        var FakeTime=sessionDurationTest+1;
```
eladBenedict: again - variable names should begin with a lowercase letter - please change all over

tests/unit/BackendClient-spec.js
```diff
@@ -0,0 +1,183 @@
...
+        clock = sinon.useFakeTimers(FakeTime);
```
eladBenedict: whenever replacing the clock, ensure to restore the real clock at the end of the test/test suite

tests/unit/BackendClient-spec.js
```diff
@@ -0,0 +1,183 @@
...
+        var BackendClient=createBackendClient(mocks);
+        BackendClient.SetLastSessionTime(0);
+        var promise = BackendClient.getLiveEntriesForMediaServer();
+
+        promise.then(function(){
+            done(new Error("Expected getLiveEntries to fail"));
```
eladBenedict: Why do you expect getLiveEntries to fail?

tests/unit/BackendClient-spec.js
```diff
@@ -0,0 +1,183 @@
...
+        }, function(){
+            done();
+        });
+    });
+    it("Calling the function .getLiveEntriesForMediaServer() twice, check that function create new session is called once!",function(done){
```
eladBenedict: Don't say what the test does, but what (functionality) it wants to check (e.g. Should not create a new session when a valid session is already available)

tests/unit/BackendClient-spec.js
```diff
@@ -0,0 +1,183 @@
...
+            './kaltura-client-lib/KalturaClient' : KalturaClientMock
+        };
+        var FakeTime=1;
+        clock = sinon.useFakeTimers(FakeTime);
+        var BackendClient=createBackendClient(mocks);
+        expect(setSessionIdMock.callCount).to.eql(0);
+        BackendClient.getLiveEntriesForMediaServer()
+            .then(function(){
+                expect(setSessionIdMock.callCount).to.eql(1);
+                clock = sinon.useFakeTimers(FakeTime+1);
```
eladBenedict: if you know that fake time is 1, I think writing 2 here would be more readable :)

tests/unit/BackendClient-spec.js
```diff
@@ -0,0 +1,183 @@
...
+        item_obj_example1={
+            dvrStatus:0,
+            id:"0_aaa"
+        };
+        item_obj_example2={
+            dvrStatus:1,
+            id:"0_bbb"
+        };
+        objsTest=[];
+        objsTest.push({"dvrWindow":100,"entryId":"0_aaa"});
```
eladBenedict: you can just create it verbatim - [{"dvrWindow":100,"entryId":"0_aaa"}, {"dvrWindow":7200,"entryId":"0_bbb"}]

tests/unit/BackendClient-spec.js
```diff
@@ -0,0 +1,183 @@
...
+        objsTest.push({"dvrWindow":7200,"entryId":"0_bbb"});
+        var results={};
+        results.objects=[];
+        results.objects.push(item_obj_example1);
```
eladBenedict: same here

tests/unit/BackendClient-spec.js
```diff
@@ -0,0 +1,183 @@
...
+            })
+        };
+        var mocks = {
+            './logger/logger': sinon.stub().returns(loggerMock),
+            './kaltura-client-lib/KalturaClient' : KalturaClientMock
+        };
+        var BackendClient=createBackendClient(mocks);
+        BackendClient.getLiveEntriesForMediaServer()
+            .then(function(objs){
+                console.log(JSON.stringify(objs));
```
eladBenedict: remove all console logging - if you want to verify something, use expect to do so (perhaps in a dedicated test)

lib/BackendClient.js
ronyad: verify 2 & 3

kaltura-hooks:
Hi @ronyad,
Thank you for contributing this pull request!
Please sign the Kaltura CLA so we can review and merge your contribution.
Learn more at http://bit.ly/KalturaContrib

kaltura-hooks:
Hi @ronyad,
Thank you for contributing this pull request!
Please sign the Kaltura CLA so we can review and merge your contribution.
Learn more at http://bit.ly/KalturaContrib

PR #53, opened 2015-11-22
update default log file names

eladBenedict:


common/config/config.json.template
```diff
@@ -11,16 +11,16 @@
   "liveDvrWindow" : 100,
   "logLevel" : "debug",
   "logToConsole" : true,
-  "logFileName" : "/var/log/dvr_worker.log",
+  "logFileName" : "/var/log/kDvrWorker.log",
```
tan-tan-kanarek: Please follow our path standards:
/opt/kaltura/log/dvr/worker.log and /opt/kaltura/log/dvr/server.log

+@Ronnysh

eladBenedict: Kaltura specific paths are committed to saas-config ; Ronny's in the loop

tan-tan-kanarek: I wasn't talking about SaaS path, the standard I mentioned is our on-prem and community standard.

PR #112, opened 2016-02-14
user is  able to view DVR content according to the entry’s configurable DVR window size

ronyad:


common/config/config.json.template
```diff
@@ -8,6 +8,7 @@
   "pollingInterval" : 2000,
   "backendPollingInterval" : 3000,
   "dvrTimeWindow" : 7200,
+  "dvrTimeWindowThreshold":900,
```
kalturaguy: put remark here what this number means

lib/BackendClient.js
```diff
@@ -79,7 +79,16 @@ var BackendClient = (function(){
                     };
 
                     if (item.dvrStatus === kalturaTypes.KalturaDVRStatus.ENABLED) {
-                        res.dvrWindow = config.get("dvrTimeWindow");
+                        var dvrWindowPartner=item.dvrWindow*60;     //   assuming dvrWindow from backend is in minutes!
```
kalturaguy: it's entryDVRWindow  not for partner

PR #114, opened 2016-02-18
Plat 5021 no smil refactor

kalturaguy:


lib/Worker.js
```diff
@@ -195,7 +195,7 @@ module.exports = (function(){
 
     var handleEntries = function(){
         var that = this;
-        return backendClient.getLiveEntriesForMediaServer().then(function(liveEntries){
+        return  adapter.getLiveEntries().then(function(liveEntries){
```
gadyaari: How do we determine the type of the adapter? Right now AdaptorFactory returns only wowzaAdaptor, but in the future?

lib/Adapters/WowzaAdapter.js
```diff
@@ -0,0 +1,91 @@
...
+                    var newEntry=entryArr[0];
+                    self.entries[entryId] = newEntry;
+
+                    newEntry.getStreamInfo=function() {
+                        return  new WowzaStreamInfo(entryId, newEntry.flavorParamsIds, stream.sourceIp);
+                    };
+
+                    return Q.resolve(newEntry);
+                });
+
```
gadyaari: Why do you return Q.resolve inside _.map() method?

lib/Adapters/WowzaAdapter.js
```diff
@@ -0,0 +1,91 @@
...
+            });
+
+           return Q.all(promises).then(function(res) {
+               return Q.resolve(res);
+           })
+
```
gadyaari: Is the "then" really necessary? Q.all returns correctly only if ALL promises are fulfilled so why do you need to call return Q.resolve as well?

PR #119, opened 2016-02-29
0.10.0 performances fixes

kalturaguy:


lib/PtsAlligner.js
```diff
@@ -48,18 +48,18 @@ var PtsAlligner = (function(){
...
         var chunkToProcess = chunkList.shift();
         var filenameToProcess = chunkToProcess.get('uri');
-        that.logger.debug("About to align PTS for " + filenameToProcess);
+        that.logger.debug("About to align PTS for %s" , filenameToProcess);
 
         var singleFileAlignmentPromise =
             qio.read(path.join(that.folderPath, filenameToProcess), "b") // Read binary data into a buffer
                 .then(function(dataBuffer) {
-                    that.logger.debug("Old PTS alignment context: " + JSON.stringify(that.context));
+                    that.logger.debug("Old PTS alignment context: %j" , that.context);
```
ronyad: Does the logger include also entry ID? not, it will be helpful to add it.

kalturaguy: It's concatenated on the beginning of each log line (together with flaovur id)

lib/FlavorDownloader.js
```diff
@@ -15,6 +15,25 @@ var util = require("util");
 var events = require("events");
 var config = require('./../common/Configuration');
 var glob = require('glob');
+var httpUtils=require('./utils/http-utils');
+
+
+function isSameM3U8(a, b) {
```
ronyad: This is a function that compare two m3u8 files?
Maybe there is "more elegant" way to check it?

kalturaguy: go a head, find one....

lib/FlavorDownloader.js
```diff
@@ -266,6 +296,12 @@ var FlavorDownloader = (function () {
                     return results;
                 }).then(function (results) {
                     return concludeOperation(results);
+                }).catch(function (err) {
+                    if (err==httpUtils.HttpDownloadNotChanged) {
+                        return Q.resolve();
```
ronyad: I thing its not good idea to write it here, since all the promises from downloadFlavorManifest() until here must not taking care in error.
I would write it like this 

        return Q.fcall(function () {
            return downloadFlavorManifest()
                .then(  #here all the logic ....#,
function(err){
 if (err instanceof  Error)
                    throw new Error("Failed to download file.\nUrl: " + url + "\nDestination: " + pathDestination + "\n" + err + " stack: " + err.stack);
                else {
                    return Q.reject(err);
})

lib/utils/http-utils.js
```diff
@@ -4,11 +4,18 @@
 
 var networkClientFactory = require('../NetworkClientFactory');
 var configuration = require('../../common/Configuration');
-var HTTPUtils = function(logger){
+var qfs = require("q-io/fs");
+var Q = require("q");
+
+
+
+var HttpDownloadNotChanged="HttpDownloadNotChanged";
```
ronyad: Maybe its better to wrap HTTPUtils module (like log decorator)  with your extension, rather than hack it.

PR #123, opened 2016-03-10
Adding method existsAndNonZero, check that file is not with zero bytes

ronyad:


lib/utils/fs-utils.js
```diff
@@ -39,10 +39,21 @@ module.exports = (function(){
             return mkdirFunc(targetPath);
         });
     };
+    var existsAndNonZero = function (path) {
+        return Q.when(qio.stat(path), function (file) {
```
kalturaguy: why do you need the Q.when?

ronyad: I actually copy paste it from q-io/fs module, that is 

    exports.exists = function (path) {
        return Q.when(this.stat(path), function () {
            return true;
        }, function () {
            return false;
        });
    };

And as it wrote in the documentation :
The when function is the static equivalent for then.

return Q.when(valueOrPromise, function (value) {
}, function (error) {
});

But I agree with you that in this case its not necessary.

PR #127, opened 2016-03-27
Plat 5040 call new api

gadyaari:


lib/BackendClient.js
```diff
@@ -128,11 +126,11 @@ var BackendClient = (function(){
         return deferred.promise;
     }
 
-    function getEntries(list) {
+    function getEntries(list, logger) {
```
kalturaguy: not sure on the order of parameters, i think logger should be first?

lib/BackendClient.js
```diff
@@ -204,33 +201,34 @@ var BackendClient = (function(){
...
             });
     };
 
-    BackendClient.getEntries = function(entryList) {
-        return createSession().
-                then(function () {
-                    return getEntries(entryList);
-                });
+    BackendClient.getEntries = function(entryList, logger) {
+        return createSession(logger)
```
kalturaguy: actully for createSession it doesn't really matter which logger it is, you can put in the general moduel logger

lib/entry/FlavorDownloader.js
```diff
@@ -83,11 +84,11 @@ var FlavorDownloader = (function() {
 
     var downloadIndefinitely = function(onIterationEndCallback) {
         var that = this;
-        that.logger.info("starting iteration");
         if (!that.shouldRun) {
```
kalturaguy: why did you delete the starting iteration log?

lib/entry/FlavorDownloader.js
```diff
@@ -83,11 +84,11 @@ var FlavorDownloader = (function() {
...
             downloaderEnd.resolve();
```
kalturaguy: please put here a log

lib/entry/FlavorDownloader.js
```diff
@@ -96,9 +97,9 @@ var FlavorDownloader = (function() {
             })
             .catch(function failure(err) {
                 that.emit("iteration-error", err);
-                that.logger.error("iteration error: %s", err);
+                that.logger.error("iteration error: %s", err.message);
```
kalturaguy: if there is an exception we will lose the call stack  (util.inspect)

lib/entry/LiveEntry.js
```diff
@@ -103,41 +93,46 @@ LiveEntry.prototype.start = function() {
...
+                    that.logger.error("Failed  to start flavor downloaders for some of the flavors:\n", errs.err);
                 }
             });
 };
 
 LiveEntry.prototype.stop = function() {
     var that = this;
+
+    if (that.FSM.current === 'stopped') {
+        return;
```
kalturaguy: put here a log telling that stop is pending

lib/utils/fsm.js
```diff
@@ -123,10 +123,10 @@
                 if (this.cannot(name))
                     return this.error(name, from, to, args, StateMachine.Error.INVALID_TRANSITION, "event " + name + " inappropriate in current state " + this.current);
```
kalturaguy: remove remarks

lib/BackendClient.js
gadyaari: If we put logger first it will cause a problem whenever someone is trying to call method in the future without passing a logger as a parameter. I think it's an option we should keep open... never know what's gonna be in the future.

lib/entry/LiveEntry.js
gadyaari: done

PR #132, opened 2016-03-31
PLAT-5040-Add-playlist-delete-fix

gadyaari:


lib/entry/LiveEntry.js
```diff
@@ -2,7 +2,7 @@
  * Created by gadyaari on 24/02/2016.
  */
 
-var masterManifestCtor = require('./../manifest/MasterManifestGenerator');
+var playlistGenerator = require('./../manifest/PlaylistGenerator');
```
kalturaguy: P and not p

PR #140, opened 2016-04-11
adding synchronized mechanism for promises

ronyad:


lib/utils/promise-utils.js
```diff
@@ -18,7 +18,32 @@ function retryPromise(fn, intervalRetry, maxRetries, errorString) {
...
+        lastApi = lastApi.finally(function() {
+
+            var deferred=Q.defer();
+            
+            var timeoutHandler= setTimeout(function(){
+                deferred.reject();
+            }, timeOut ? timeOut : defultTimeOut);
+
+            fn().then(function (){
+                clearTimeout(timeoutHandler);
```
kalturaguy: What if you will get a resolve after timeout....?

lib/utils/draft.js
```diff
@@ -0,0 +1,3 @@
+/**
```
kalturaguy: Ma ze?

lib/utils/promise-utils.js
```diff
@@ -18,7 +18,32 @@ function retryPromise(fn, intervalRetry, maxRetries, errorString) {
     });
 }
 
+function SynchornizedPromises() {
+    var lastApi =  Q.resolve();
+    var defultTimeOut = 20000; //~ 20 seconds
```
kalturaguy: Increase to 1 min. Just in case...

lib/utils/promise-utils.js
kalturaguy: Increase to 1 min. Just in case... maybe pass it via ctor

PR #142, opened 2016-04-17
Adding the module name to the log file

ronyad:


lib/BackendClient.js
```diff
@@ -9,10 +9,9 @@ var kalturaClient = require('./kaltura-client-lib/KalturaClient');
 var kalturaClientBase = require('./kaltura-client-lib/KalturaClientBase');
 var kalturaTypes = require('./kaltura-client-lib/KalturaTypes');
 var config = require('./../common/Configuration');
-var moduleLogger = require('./logger/logger')(module);
-
+var logger = require('../common/logger/loggerNew').getLogger('BackendClient');
```
kalturaguy: rename to moduleLogger and you won't need all the changes below....

dvr-ws/logger/logger.js
```diff
@@ -2,8 +2,57 @@
  * Created by Elad Benedict on 8/24/15.
  */
```
kalturaguy: why aren't we using the same logger as in common?

lib/manifest/TwoPhasedChunklistManifestGenerator.js
```diff
@@ -18,10 +18,11 @@ module.exports = function(downloadedChunksGetter, outputFolderPath, filename, ti
     // Once chunks are available they are exposed through the externalized chunklist manifest
 
 
-    logger = loggerDecorator(logger, "Two phased manifest generator: ");
```
kalturaguy: put a shorter name and in one world, category must be a one word

lib/utils/promise-utils.js
```diff
@@ -1,7 +1,7 @@
 /**
  * Created by gadyaari on 07/02/2016.
  */
-var logger = require('./../logger/logger')(module);
+var logger = require('../../common/logger').getLogger('promise-utils');
```
kalturaguy: do you use this logger?, if yes, then you have duplicated name with SynchornizedPromises

PR #149, opened 2016-04-26
SUP-8067 Update playlist bitrate during stream

gadyaari:


common/config/config.json.template
```diff
@@ -17,7 +17,7 @@
   "sessionDuration" : 180000,
   "apiUpdateInterval" : 60000,
   "flavorDownloaderTeardownInterval" : 15000,
-  "playlistUpdateInterval" : 300000,
+  "playlistUpdateInterval" : 5,
```
kalturaguy: please be consistent if other are in ms then this needs to be too

lib/entry/StateManager.js
```diff
@@ -77,14 +80,16 @@ module.exports = function(entryObject, playlistGenerator) {
...
+        if (from === "suspending"){
             logger.warn("Something went wrong:" + stateMachineMessagePattern, event.toUpperCase(), from, to)
         }
         var currTime = new Date().getTime();
         if ((lastServerUpdate === null) || ((currTime - lastServerUpdate) >= serverUpdateInterval)) {
-            // When reporting play update entry's playlist
-            playlistGenerator.updatePlaylist();
             logger.info(stateMachineMessagePattern, event.toUpperCase(), from, to);
+            // When reporting play update entry's playlist very 5 min
+            if (!(shouldUpdatePlaylist++ % config.get('playlistUpdateInterval'))) {
```
kalturaguy: doesn't look nice 👎

lib/manifest/PlaylistGenerator.js
```diff
@@ -78,15 +77,16 @@ function savePlaylist() {
...
-    var currTime = new Date();
     // Update the entry's playlist every minute
-    if ((currTime - that.lastPlaylistUpdate) > config.get('playlistUpdateInterval')) {
-        that.logger.info("Updating playlist");
-        return savePlaylist.call(this)
-            .catch(function(error) {
-                that.logger.error("Unable to update playlist: %s", error);
-            });
-    }
+    that.logger.debug("Retrieving flavors for update");
```
kalturaguy: question: what happens if you call updatePlaylist, then the stream stops, and then the getAllFlavours return?

lib/entry/StateManager.js
```diff
@@ -77,14 +79,17 @@ module.exports = function(entryObject, playlistGenerator) {
...
+        if (from === "suspending"){
             logger.warn("Something went wrong:" + stateMachineMessagePattern, event.toUpperCase(), from, to)
         }
         var currTime = new Date().getTime();
         if ((lastServerUpdate === null) || ((currTime - lastServerUpdate) >= serverUpdateInterval)) {
-            // When reporting play update entry's playlist
-            playlistGenerator.updatePlaylist();
             logger.info(stateMachineMessagePattern, event.toUpperCase(), from, to);
+            // When reporting play update entry's playlist very 5 min
+            if (shouldUpdatePlaylist % config.get('playlistUpdateTimeInMin') === 0) {
```
kalturaguy: first time will call updatePlaylist twice since shouldUpdatePlaylist == 0

kalturaguy: playlistUpdateTimeInMin?  but this is multiply by serverUpdateInterval... so if serverUpdateInterval is 12000 then it's two minutes

PR #151, opened 2016-04-27
PLAT-5421 Gix bug while failing to stop entry

gadyaari:


lib/entry/LiveEntry.js
```diff
@@ -117,17 +117,18 @@ LiveEntry.prototype.stop = function() {
...
-                    });
+                    return stopPromise[1]
+                        .catch(function() {
+                            return Q.resolve();
+                        });
                 })
                 .catch(function(err) {
-                    that.logger.error("Failed stopping flavor downloader, error: %s", err);
-                    throw err;
+                    that.logger.error("Error stopping entry: %s", err.message);
```
kalturaguy: do $j and err instead of err.message

PR #154, opened 2016-04-29
Add x kaltura session to the logging of BackEndClient

ronyad:


lib/BackendClient.js
```diff
@@ -49,7 +49,7 @@ var BackendClient = (function(){
             client.session.start(function(results, err, headers) {
                 var endTime  = new Date();
                 var operationTime = endTime - now;
-                logger.debug("Request for createSession took %s", operationTime);
+                logger.debug("Request for createSession took %s, %s", operationTime, headers[7]);
```
yossipapi: Taking the index from the headers is not correct. 
For example if the request has returned from the cache you won't have x-Kaltura-session on the response headers.
Another thing is that the session without the Machine x-me would not be very helpful.

I think the proper thing would be to print out all the headers of the response.

Same applies for all the other changes as well.

PR #156, opened 2016-05-01
New ts rebase

ronyad:


common/config/config.json.template
```diff
@@ -2,6 +2,7 @@
   "mockNetwork" : false,
   "mockBackend" : false,
   "readMockBackendResponseFromFile" : false,
+  "universalTsTimestamp" : true,
```
kalturaguy: can be deleted, nobody uses it

PR #167, opened 2016-05-08
Crete new error in case that promise is rejected

ronyad:


lib/Worker.js
```diff
@@ -74,8 +74,8 @@ module.exports = (function(){
                             return liveEntry;
                         })
                         .catch(function(err) {
-                            err.entry = liveEntry;
                             logger.debug("Live entry [%s] Failed to start", e.entryId);
```
kalturaguy: it was helpfull if you would print err....

lib/entry/LiveEntry.js
```diff
@@ -138,7 +138,8 @@ LiveEntry.prototype.stop = function() {
                 });
         })
         .catch(function() {
-            return Q.reject();
+            //todo we need this line? ron 
+            return Q.reject(new Error("Failed to stop promise"));
```
kalturaguy: why not pass the error in the catch?

lib/utils/promise-utils.js
```diff
@@ -37,7 +37,7 @@ function SynchornizedPromises(logger) {
                         deferred.resolve(result[0].value);
                     }
                     else {
-                        deferred.reject(result[0].reason);
+                        deferred.reject(new Error ("result[0].reason"));
```
kalturaguy: deferred.reject(result[0].reason); will reject the actual error

lib/utils/error-utils.js
```diff
@@ -8,6 +8,9 @@ var ErrorUtils = (function(){
 
     var obj = {};
 
+    obj.printError=function (err) {
```
kalturaguy: rename to error2string

lib/Worker.js
```diff
@@ -74,8 +75,10 @@ module.exports = (function(){
                             return liveEntry;
                         })
                         .catch(function(err) {
-                            err.entry = liveEntry;
-                            logger.debug("Live entry [%s] Failed to start", e.entryId);
+                            logger.error("Live entry [%s] Failed to start", e.entryId);
+                            if (_.isObject(err)){
```
gadyaari: Please move [%s] to beginning of string to be consistent with other logger printings
,

PR #178, opened 2016-05-25
Support cases where entry server node is not yet available on entry st…

yossipapi:
…artup

### DIFF
--- a/lib/BackendClient.js
+++ b/lib/BackendClient.js
@@ -239,7 +239,7 @@ var BackendClient = (function(){
     function updateStreamInfo(entryId, entryServerNodeId, liveEntryServerNode)
     {
         var deferred = Q.defer();
-        logger.debug("[%s] Calling updateStreamInfo for entryServerNode id [%s], with stream info array [$j]",entryId,  entryServerNodeId, liveEntryServerNode.streams);
+        logger.debug("[%s] Calling updateStreamInfo for entryServerNode id [%s], with stream info array [%j]",entryId,  entryServerNodeId, liveEntryServerNode.streams);
 
         client.entryServerNode.update(function(result, err, headers) {
             printAPIResponse(result, headers);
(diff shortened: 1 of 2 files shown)


lib/BackendClient.js
```diff
@@ -252,6 +252,43 @@ var BackendClient = (function(){
         return deferred.promise;
     }
 
+    function getLiveEntryServerNodeId(entryId, serverType) {
```
kalturaguy: why by serverType?  why not by the serverId?

lib/manifest/PlaylistGenerator.js
```diff
@@ -53,6 +54,13 @@ function checkTranscodingProfileChange() {
         });
 }
 
+function updateStreamInfo() {
+    backendClient.updateStreamInfo(this.entryId, this.entryServerNodeId, this.flavorsObjArray)
```
kalturaguy: forgot to add return

yossipapi: Added

lib/BackendClient.js
yossipapi: What is the difference? Why is using one better than the other?

lib/manifest/PlaylistGenerator.js
```diff
@@ -73,10 +81,16 @@ function savePlaylist() {
         .then(function() {
             that.lastPlaylistUpdate = new Date();
             that.logger.debug("Playlist written to disk");
-            backendClient.updateStreamInfo(that.entryId, that.entryServerNodeId, that.flavorsObjArray)
-                .then(function() {
-                    that.logger.debug("Flavors updated in DB");
-                });
+            if(that.entryServerNodeId) {
```
kalturaguy: is this nicer?
 var getLiveEntryServerNodeIdPromise=that.entryServerNodeId ? Promise.resolve(that.entryServerNodeId) :  backendClient.getLiveEntryServerNodeId(that.entryId, that.entryServerType);

getLiveEntryServerNodeIdPromise.then(function(entryServerNodeId) {
                         that.entryServerNodeId = entryServerNodeId;
                         updateStreamInfo.call(that);
                     });

yossipapi: Yep looks better for the Promise lovers among us :)
I will update the code accordingly.
👍

lib/manifest/PlaylistGenerator.js
```diff
@@ -73,10 +83,16 @@ function savePlaylist() {
         .then(function() {
             that.lastPlaylistUpdate = new Date();
             that.logger.debug("Playlist written to disk");
-            backendClient.updateStreamInfo(that.entryId, that.entryServerNodeId, that.flavorsObjArray)
-                .then(function() {
-                    that.logger.debug("Flavors updated in DB");
-                });
+
+            var getLiveEntryServerNodeIdPromise =
+                that.entryServerNodeId ? Q.Promise.resolve(that.entryServerNodeId)
```
kalturaguy: does this work?  i know only Q.resolve not Q.Promise.resolve

PR #179, opened 2016-05-31
reset entry info when reciving Wowza connection error

yossipapi:


lib/Adapters/WowzaAdapter.js
```diff
@@ -49,7 +49,8 @@ WowzaAdapter.prototype.getLiveEntries = function() {
     var that = this;
     return networkClient.read({
             url : wowzaUrl,
-            json : true
+            json : true,
+            timeout: 20000
```
kalturaguy: we should use 
config.get("requestTimeout")

(already in config.json)

yossipapi: Done please review

PR #199, opened 2016-06-16
0.13.0 plat 5120

lilachmaliniak:


common/config/config.json.template
```diff
@@ -34,13 +35,15 @@
     "port" : 1935,
     "user" : "",
     "password" : ""
+
   },
   "webServerParams" : {
```
kalturaguy: this section should be gone

kalturaguy:
i recommend to re-write it and not merge,

PR #201, opened 2016-06-19
1.0.0 0.14merge

kalturaguy:


lib/BackendClient.js
```diff
@@ -40,18 +44,26 @@ var BackendClient = (function(){
...
 
     function createSession() {
-        var deferred = Q.defer();
+
         var now = new Date();
+
         //create session
-        if (!lastSessionTime || (now - lastSessionTime) > sessionDuration) {
+        if (!sessionRequestPromise || !lastSessionTime || (now - lastSessionTime) > sessionDuration) {
+
```
gadyaari: !sessionRequestPromise || is not really necessary (but also not a big issue if it stays)

PR #215, opened 2016-07-07
1.0 plat 5586 bis

igorshevach:


igorshevach:
- remove ffmpeg warnings during ts to mp4 conversion
- sync starting offsets between different flavors/tracks to reference same timestamp (pts) since video dts times aren't accurate for that purpose.

PR #220, opened 2016-07-14
01.0 plat 5802

igorshevach:


igorshevach:
adjust logging levels to prevent flooding the logs

PR #224, opened 2016-07-18
Regression test implementation tested for single flavor

lilachmaliniak:
Regression test is implemented for multiple flavors but currently tested, rouphly, for single.
work in progress to fully support multiple flavours.
Known issue: 
chunklist checksums do not match between ground truth and test results.


lilachmaliniak:
pull opened to wrong branch

PR #225, opened 2016-07-18
Regression test implementation tested for single flavor

lilachmaliniak:
Regression test is implemented for multiple flavors but currently tested, rouphly, for single.
work in progress to fully support multiple flavours.
Known issue: 
chunklist checksums do not match between ground truth and test results.


lib/Adapters/RegressionAdapter/hlsStreamRegressionValidator.js
```diff
@@ -0,0 +1,231 @@
...
+    return this.loadResultsDatabase()
+        .then( function() {
+            that.logger.debug('[%s] regression tests validator successfully initialized.', that.entryId);
+        })
+        .catch (function(err) {
+            that.logger.error('[%s] failed to initialize regression tests validator. Error: %s, stack: %s', that.entryId, err.message, err.stack);
+        });
+}
+
+HlsStreamRegressionValidator.prototype.createDestinationDir = function() {
```
kalturaguy: should +HlsStreamRegressionValidator.prototype.createDestinationDir  be in utils?

lib/Adapters/RegressionAdapter/validatorUtility.js
```diff
@@ -0,0 +1,83 @@
...
+       else if (!valid) {
+           diff['result'] = 'regression test failed';
+       }
+       else {
+           diff['result'] = 'validation skipped. config \'override\' value=true';
+       }
+       var result_path = that.config.path;
+       result_path += '/' + that.entryId + '_regression_test_result.json';
+       // save diff
+       fs.writeFile(result_path, JSON.stringify(diff, null, 2), function (write_err) {
```
kalturaguy: use q-io writeFile instead

lib/Adapters/RegressionAdapter/hlsStreamRegressionValidator.js
```diff
@@ -0,0 +1,231 @@
...
+
+            // todo: count files and delete old files so that total number including new one is up to history_count
+
+            if (that.override === true) {
+                that.logger.debug("[%s] creating new database for regression test results!", that.entryId);
+            }
+            else {
+                that.logger.debug('Going to open %s. Override=%s', that.fullpath, that.override);
+
+                fs.readFile(that.fullpath, 'utf8', function (err, data) {
```
kalturaguy: use q-io readFile

lib/Adapters/RegressionAdapter/hlsStreamRegressionValidator.js
```diff
@@ -0,0 +1,231 @@
...
+
+HlsStreamRegressionValidator.prototype.saveResultsToFile = function (reason) {
+
+    var that = this;
+    var deferred = Q.defer();
+
+    this.logger.info('[%s] saving regression results. Reason: %s', this.entryId, reason);
+
+    if (this.override || this.regression_results_db === undefined) {
+        fs.writeFile(this.fullpath, JSON.stringify(this.this_run_results, null, 2), function (err) {
```
kalturaguy: use qio writefile

lib/Adapters/RegressionAdapter/RegressionAdapter.js
```diff
@@ -0,0 +1,445 @@
...
+        var mediaServerPlaylistUrl = that.getFullLiveUrl(entryConfig.entryId, 'all');
+        var fullUrl = url.resolve(mediaServerPlaylistUrl, 'playlist.m3u8');
+
+        // read and parse the playlist and assemble playlist per flavor
+        return networkClient.read(fullUrl)
+            .then(function (content) {
+                that.initPlaylistPerFlavor(entryConfig.entryId, content);
+            })
+            .catch(function (error) {
+                that.logger.warn('Initialization of entryId failed with error %s. Failed to prepare playlist per flavor.', error);
```
kalturaguy: what about callstack

lib/Adapters/RegressionAdapter/RegressionAdapter.js
```diff
@@ -0,0 +1,445 @@
...
+    }
+
+    return playlist;
+}
+
+TestHelper.prototype.getLastChunklistFileIndex = function (entryId, fullPath) {
+
+    var that = this;
+
+    if (!String.prototype.endsWith) {
```
kalturaguy: can be removed now with ES6

lib/Adapters/RegressionAdapter/RegressionAdapter.js
```diff
@@ -0,0 +1,445 @@
...
+            }
+            position -= searchString.length;
+            var lastIndex = subjectString.indexOf(searchString, position);
+            return lastIndex !== -1 && lastIndex === position;
+        };
+    }
+
+    var deferred = Q.defer();
+
+    fs.readdir(fullPath, function (err, items) {
```
kalturaguy: there is already an implementation (glob?)

lib/Adapters/RegressionAdapter/hlsStreamRegressionValidator.js
lilachmaliniak: done, replaced with qio.write(...)

lib/Adapters/RegressionAdapter/hlsStreamRegressionValidator.js
lilachmaliniak: no need to add to utils. Replaced with call to q-io makeTree().

lilachmaliniak: no need to add to utils. Replaced with call to q-io makeTree().

lib/Adapters/RegressionAdapter/hlsStreamRegressionValidator.js
lilachmaliniak: done.

lib/Adapters/RegressionAdapter/hlsStreamRegressionValidator.js
lilachmaliniak: done.

lib/Adapters/RegressionAdapter/RegressionAdapter.js
lilachmaliniak: replaced in all places where error is printed to log with ErrorUtils.error2string(err).

lib/Adapters/RegressionAdapter/RegressionAdapter.js
lilachmaliniak: removed.

lilachmaliniak:
finished implementing all review comments.

lib/Adapters/RegressionAdapter/RegressionAdapter.js
```diff
@@ -0,0 +1,379 @@
...
+
+TestHelper.prototype.initEntries = function() {
+
+    var that = this;
+    var root_mp4_path = config.get('rootFolderPath');
+    var p;
+    p = _.map(entriesInfo, function (entryConfig) {
+        // remove directory
+        var fullpath = root_mp4_path + '/' + entryConfig.entryId;
+        return fs_utils.cleanFolder(fullpath, 'obsolete').then(function () {
```
kalturaguy: cleanFolder of fs_utils is not really cleaning the folder, but it only renames it to another location (performance)
suggest you use https://www.npmjs.com/package/q-io#removetreepath

lilachmaliniak: done

lib/Adapters/AdapterFactory.js
```diff
@@ -2,15 +2,22 @@
...
 var config = require('../../common/Configuration');
 
 var simulateStreams = config.get('simulateStreams');
 var regressionAdapterConfig = config.get('regressionAdapter');
+var command = process.argv.length > 2 ? process.argv[2] : undefined;
+
 
 exports.getAdapter = function() {
 
+    if (command != undefined) {
```
kalturaguy: this doesn't belong here  if you want to override configuration it should be in a code that handles config (maybe config.js...)

package.json
```diff
@@ -2,10 +2,10 @@
...
   "version": "1.0.0",
   "repository": {
-      "type": "git",
-      "url": "https://github.com/kaltura/liveDVR"
-    },
-  "license": "",
+    "type": "git",
+    "url": "git+https://github.com/kaltura/liveDVR.git"
+  },
+  "license": "ISC",
```
kalturaguy: why ISC? is this aligned with other products in kaltura?

package.json
```diff
@@ -22,22 +22,37 @@
...
-    "chai-as-promised":"4.1.1"
+    "chai-as-promised":"4.1.1",
+    "grunt-execute": "^0.2.2"
   },
   "dependencies": {
     "curry": "1.2.0",
     "forever": "0.15.2",
     "glob": "7.0.5",
     "log4js": "0.6.37",
+    "m3u8": "git://github.com/eladBenedict/node-m3u8.git#master",
```
kalturaguy: why did you re-add it?

package.json
lilachmaliniak: I don't know. This property was added after automatic init to add grunt-execute.

lilachmaliniak: removed.

lib/Adapters/RegressionAdapter/hlsStreamRegressionValidator.js
```diff
@@ -0,0 +1,254 @@
...
+const ErrorUtils = require('../../utils/error-utils');
+
+function checksum (str, algorithm, encoding) {
+    return crypto
+        .createHash(algorithm || 'md5')
+        .update(str, 'utf8')
+        .digest(encoding || 'hex')
+}
+
+let logheader1 = function(literals, ...values) {
```
kalturaguy: this is duplicate code, the logger has already a build in feature of this, please use it

package.json
```diff
@@ -36,8 +36,21 @@
...
     "underscore": "1.8.3",
     "v8-profiler": "5.6.5"
-  }
+  },
+  "description": "DVR implementation built on top of live streams",
+  "bugs": {
+    "url": "https://github.com/kaltura/liveDVR/issues"
+  },
+  "homepage": "https://github.com/kaltura/liveDVR#readme",
+  "main": "grunt-config.js",
```
kalturaguy: why the main entry is the grunt?

lilachmaliniak: This is a mistake. I change it to lib/App.js

common/config/config.json.template
```diff
@@ -23,8 +23,29 @@
...
   "monitorPort": 4390,
+  "preserveOriginalHLS": false,
+  "recordingDurationMinutes": 5,
+  "overrideOldRecordedHLS": true,
   "regressionAdapter": {
-      "enable":false
+      "enable":false,
+      "analysis": false,
+      "analyzerRootPath" : "TBD",
+      "dataWarehouseRootFolderPath" : "instructions: put here root path to data warehouse (e.g. /var/tmp/regression/data_warehouse)",
```
kalturaguy: why not use ~/.....

package.json
```diff
@@ -36,8 +38,22 @@
     "request": "^2.61.0",
     "shelljs": "0.7.0",
     "touch": "1.0.0",
-    "chunked-stream": "0.0.2",
     "underscore": "1.8.3",
-    "v8-profiler": "5.6.5"
-  }
+    "v8-profiler": "5.6.5",
+    "child_process": "1.0.2"
+  },
```
kalturaguy: you removed chunked-stream, but it's still used

package.json
```diff
@@ -36,8 +38,22 @@
     "request": "^2.61.0",
     "shelljs": "0.7.0",
     "touch": "1.0.0",
-    "chunked-stream": "0.0.2",
     "underscore": "1.8.3",
-    "v8-profiler": "5.6.5"
-  }
+    "v8-profiler": "5.6.5",
+    "child_process": "1.0.2"
```
kalturaguy: please don't use this package, node.js has already all functionality you need

lib/utils/general-utils.js
```diff
@@ -0,0 +1,30 @@
+/**
```
kalturaguy: don't think this belong here, this is used only by ControllerWrapper, so it should be there

lib/Controller.js
```diff
@@ -1,7 +1,8 @@
 /**
  * Created by elad.benedict on 8/26/2015.
  */
-
+// parse command line args and update config obj
+var regressionConfig = require('./Adapters/RegressionAdapter/RegressionConfig.js').getConfig();
```
kalturaguy: this shouldn't belong here, the Controller shouldn't know what regressionAdapter is

lib/Controller.js
```diff
@@ -226,16 +235,33 @@ module.exports = (function(){
...
+                // Todo: check if a bug --> stop() doesn't work as expected
+                // that.stop();
+            }
+        });
+        adapter = adapter.getAdapter(that);
+
         logger.warn("Starting liveProxy pid: %s, wowza: '%s', API server: '%s', prefix='%s'",process.pid,config.get("mediaServer").hostname,config.get('backendClient').serviceUrl,that.prefix);
         logger.warn("Log files %s",config.get('logFileName'));
         return Q.fcall(function() {
+            if (regressionConfig.command && regressionConfig.command.record ||
```
kalturaguy: please move it from here... (same comment from above) you already have Controller wrapper

lib/Adapters/TestAdapter.js
```diff
@@ -10,6 +10,7 @@ var networkClient = require('./../NetworkClientFactory').getNetworkClient();
 var url=require('url');
```
kalturaguy: why did you need to change this file?

lib/Adapters/RegressionAdapter/hlsStreamRegressionValidator.js
```diff
@@ -0,0 +1,296 @@
...
+
+    var that = this;
+    var deferred  = Q.defer();
+
+    data.errors[data.flavor] = {errors: {}};
+    var flavor_header = logheader1 `${this.entryId}${data.flavor}${data.index}`;
+
+    let url = util.format("http://localhost:8080/kLive/smil:%s_all.smil/%s/%s", this.entryId, data.flavor, item.properties.uri);
+
+    networkClient.read({url: url, timeout: 10000})
```
kalturaguy: you are downloading the entire file just to know the length....
i suggest you do a HEAD http request which will return the length and not the content

lib/Adapters/RegressionAdapter/StreamRegressionValidatorFactory.js
```diff
@@ -0,0 +1,23 @@
...
+var hlsStreamRegressionValidator = require('./hlsStreamRegressionValidator');
+var hlsAnalysisStreamRegressionValidator = require('./hlsAnalysisStreamRegressionValidaotor');
+var analysis = config.get('regressionAdapter').analysis;
+
+var getValidator = function(entryConfig, flavors, last_chunklist_index) {
+
+    if (analysis) {
+        return new hlsAnalysisStreamRegressionValidator(entryConfig, flavors, last_chunklist_index);
+    } else {
+        return new hlsStreamRegressionValidator(entryConfig, flavors, last_chunklist_index);
```
kalturaguy: the name is not good... maybe HlsChecksumRegressionValidator ?

common/config/config.json.template
lilachmaliniak: replaced with "~/live-monitor/"

package.json
lilachmaliniak: added chunked-stream and removed child_process.

package.json
lilachmaliniak: ok - code will change

lib/utils/general-utils.js
lilachmaliniak: removed

package.json
lilachmaliniak: using node original child_process.

lib/Controller.js
lilachmaliniak: moved to AdapterFactory.js. Controller code cleaned from unit tests related code.

lib/Adapters/TestAdapter.js
lilachmaliniak: 1) support multiple different recording of same entry, based on override flag.
2) fixed the naming convention of recorded stream output directory to avoid using spaces in name.
3) added timer to limit recording time based on configuration.

lib/Adapters/RegressionAdapter/hlsStreamRegressionValidator.js
lilachmaliniak: done. Now request options obj include method: HEAD.

lib/Adapters/RegressionAdapter/StreamRegressionValidatorFactory.js
lilachmaliniak: replaced. did more renaming:
hlsStreamRegressionValidator --> hlsChunklistRegressionValidator  StreamRegressionValidatorFactory --> RegressionValidatorFactory 
StreamRegressionValidatorBase --> RegressionValidatorBase  hlsAnalysisStreamRegressionValidator --> hlsAnalysisRegressionValidator

lib/Adapters/RegressionAdapter/hlsChecksumRegressionValidator.js
```diff
@@ -0,0 +1,300 @@
...
+    //let url = util.format("http://localhost:8080/kLive/smil:%s_all.smil/%s/%s", this.entryId, data.flavor, item.properties.uri);
+    let options = {
+        method: 'HEAD',
+        url: `http://localhost:8080/kLive/smil:${this.entryId}_all.smil/${data.flavor}/${item.properties.uri}`,
+       
+        timeout: 10000};
+    networkClient.read(options)
+        .then(function (content) {
+            // add to results new segment
+            var ts_id = item.properties.uri.split('.')[0];
```
kalturaguy: if you are doing head request then what is content now?  content.length will probably be something else... you need the header 'content-length'

lib/Adapters/RegressionAdapter/hlsChecksumRegressionValidator.js
```diff
@@ -0,0 +1,300 @@
...
+            if (!_.has(that.this_run_results[data.flavor].ts[ts_id])) {
+                that.this_run_results[data.flavor].ts[ts_id] = content.length;
+                deferred.resolve();
+            } else if (that.this_run_results[flavorId].ts[ts_id] != content.length) {
+                that.logger.debug('%s %s validation failed', flavor_header, ts_id);
```
kalturaguy: instead of  passing the flavor_header on each log line, it's better if you have created a new logger with this as prefix, see example in:

https://github.com/kaltura/liveDVR/blob/1.0/lib/entry/FlavorDownloader.js#L55

lib/Adapters/RegressionAdapter/hlsChecksumRegressionValidator.js
lilachmaliniak: done. The length is taken from content-length header filed. It required also update of NetworkClient to retune respose.headers instead of body, when method: HEAD is part of request.

gruntfile.js
```diff
@@ -11,9 +11,13 @@ module.exports = function (grunt) {
...
 
+    grunt.loadNpmTasks('grunt-execute');
+
     grunt.loadTasks('./lib/grunt/tasks');
     grunt.loadTasks('./lib/grunt/config');
 
+
     // Default task(s).
-    grunt.registerTask('default', ['jshint:dev','unit-test:dev', 'component-test:dev', 'mocha_istanbul:dev']);
+   // grunt.registerTask('default', ['jshint:dev', 'execute:regression_test', 'unit-test:dev', 'component-test:dev', 'mocha_istanbul:dev']);
```
kalturaguy: is now the unit-test disabled?

common/config/config.json.template
```diff
@@ -23,8 +23,29 @@
   "logFileName" : "~/dvr/kLiveController.log",
   "logTimeZoneOffset" : 300,
   "monitorPort": 4390,
+  "preserveOriginalHLS": false,
+  "recordingDurationMinutes": 5,
```
kalturaguy: can you please move that to regessionAdapter or such?  it will be confusing for the future when we will do the real live to vod recording in the live controller

lib/NetworkClient.js
```diff
@@ -19,7 +19,11 @@ function readRequest(requestData)
             return;
         }
         if (!error && response.statusCode === 200) {
-            d.resolve(body);
+            if (!requestData.method || requestData.method != 'HEAD') {
```
kalturaguy: don't like this much.... 
maybe add resolveWithFullResponse to the requestData and if this true then return
d.resolve ({body,response.headers})
(so we can reuse this)

lib/Controller.js
```diff
@@ -253,14 +259,18 @@ module.exports = (function(){
...
-
+    
     Controller.prototype.getStat = function() {
         return {
             "handledEntries": this.handledEntries,
             "phantomEntries": this.phantomEntries
         }
     };
 
+    Controller.prototype.getAdapter = function() {
```
kalturaguy: why exposing adapter with the controller?  you can call 
var adapter = require('./Adapters/AdapterFactory.js').getAdapter();
anywhere in code...

lib/Controller.js
```diff
@@ -15,8 +16,10 @@ var Q = require('q');
 var qio = require('q-io/fs');
 var path = require('path');
 var util = require('util');
-var adapter = require('./Adapters/AdapterFactory.js').getAdapter();
+
 var MonitorServer=require('./MonitorServer.js');
+const events = require('events');
+
```
kalturaguy: nothing in this file really change, so for each merge revert all space changes so this file will not be in the pull reuqest

lib/Controller.js
lilachmaliniak: removed getAdapter from Controller.js and updated AdapterFactory to support only one timer adapter instantitiation.

lib/Controller.js
lilachmaliniak: Guy, there is one change that is required, unless I add the line to App.js...
var adapter = require('./Adapters/AdapterFactory.js').getAdapter(); moved to the top of Controller.js so the configuration updates will take place before any module get config.

lilachmaliniak: Guy,
there is one change that is required, unless I add the line to App.js...
var adapter = require('./Adapters/AdapterFactory.js').getAdapter();
moved to the top of Controller.js so the configuration updates will take place before any module get config.

PR #233, opened 2016-08-09
- PLAT-5939 unexpectedly large pts to dts delay  

igorshevach:
- PLAT-5939. added diagnostic messages from converter
- setenv.sh debug/release build fix
- json serialization fix when there are no keyFrameDurations present (e.g audio only)


common/Configuration.js
```diff
@@ -21,6 +21,7 @@ module.exports = (function(){
     if (fs.existsSync(mappingFilePath))
     {
         var mappingContent = fs.readFileSync(mappingFilePath, 'utf8');
+        mappingContent= mappingContent.replace(/~/g,hostname.homedir());
```
kalturaguy: maybe it's better to move that ourside the scope  and remove the one line line#16??

igorshevach: agreed

PR #239, opened 2016-08-24
PLAT-5840: Create file hierarchy

gadyaari:


lib/playlistGenerator/ConcatSource.js
```diff
@@ -55,6 +55,12 @@ function ConcatSource(loggerInfo, playlistObj, parentObj, track, chunkPath) {
 
 }
 
+function retractBasePath(path) {
```
igorshevach: please check PlaylistGenerator.checkFileExists and PlaylistGenerator.handleClipError are not broken
generally, track all usages of path.basename related to chunk path under playlistGenerator dir and see if breaks existing code. maybe you'd better move  retractBasePath to persistenceFormat.js to be globally accessible...

common/PersistenceFormat.js
```diff
@@ -14,8 +14,9 @@ module.exports = {
         return path.join(config.get('rootFolderPath'), entryId);
     },
```
kalturaguy: i think the file name (PersistenceFormat.js) needs a better naming...

common/PersistenceFormat.js
```diff
@@ -14,8 +14,9 @@ module.exports = {
...
 
-    getEntryRelativePath: function (entryId) {
-        return entryId;
+    getEntryBasePath: function (path) {
+        let subString = path.match(/([01]_\w+\/[0-9]+\/)/)[1];
+        return path.substr(0, path.lastIndexOf(subString) + subString.length);
     },
```
kalturaguy: not sure what are you trying to get here, also your regex needs to be modified

common/PersistenceFormat.js
```diff
@@ -33,9 +34,12 @@ module.exports = {
     getMasterManifestName: function () {
         return 'playlist.json';
     },
+    
     getMP4FileNamefromInfo: function(chunkPath){
         return path.basename(chunkPath) + '.mp4';
     },
```
kalturaguy: we need to minimze the file name length by at least remove the .ts

common/PersistenceFormat.js
```diff
@@ -33,9 +34,12 @@ module.exports = {
...
+    
     getTSChunknameFromMP4FileName: function(mp4FileName){
-        return mp4FileName.substr(0,mp4FileName.length -'.mp4'.length);    }
+        return mp4FileName.substr(0,mp4FileName.length -'.mp4'.length);
```
kalturaguy: '.mp4'.length will always be 4 (optimization)

lib/entry/FlavorDownloader.js
```diff
@@ -119,41 +120,43 @@ var FlavorDownloader = (function() {
...
-                    return mp4File;
-                })
-                .catch(function(err){
-                    // do not postpone download of the next chunks in line when chunk is corrupted
-                    if(err instanceof MP4ConversionError){
-                        that.logger.warn("failed to convert: error string: %s, chunk: %s", err,err.chunkName);
-                        delete that.chunksToDownload[chunkName];
-                        return Q.resolve(null);
-                    }
+            return fsUtils.createHierarchyPath(that.destFolderPath)
```
kalturaguy: performance issue? for every .ts we call maketree, is this OK?

common/PersistenceFormat.js
gadyaari: Agreed. What should we change it to?

common/PersistenceFormat.js
gadyaari: When PlaylistGenerator creates a new ConcatSource instance it send as a parameter the path for the folder where to store the files (retrieved from fs-utils):
/web/content/kLive/live/<entry_Id>/<day>/<hour>/ 
Since the ConcatSource requires only the BASE path I need to attract it from the full file path. That way the playlist.json would show as basePath: /web/content/kLive/live/<entry_Id>/
and each file specific path would be: <day>/<hour>/<file_name>/

Why do I need to modify the regEx? it works fine

common/PersistenceFormat.js
gadyaari: Haven't touched the file name. I assumed it is a necessity of Igor

common/PersistenceFormat.js
gadyaari: A function I'm not familiar with. Igor uses it, will check with him

lib/entry/FlavorDownloader.js
gadyaari: Was thinking about it when wrote it. Still testing to see performance results. In general, makeTree only completes the paths that do not exist, so there's no recreation of the directory with each call.

common/PersistenceFormat.js
```diff
@@ -10,32 +10,44 @@ var Q = require('q');
```
kalturaguy: need to rename file

common/PersistenceFormat.js
```diff
@@ -10,32 +10,42 @@ var Q = require('q');
...
 
-    getEntryFullPath: function (entryId) {
+    getEntryBasePath: function (entryId) {
         return path.join(config.get('rootFolderPath'), entryId);
     },
 
-    getEntryRelativePath: function (entryId) {
-        return entryId;
+    getBasePathFromFull: function (path) {
+        let subString = path.match(/([01]_\w+\/[0-9]+\/)/)[1];
```
igorshevach: maybe it is better to keep precompiled regex object  /([01]_\w+\/[0-9]+\/)/ instead

PR #243, opened 2016-08-30
1.0 plat 6001

igorshevach:


lib/playlistGenerator/Playlist.js
```diff
@@ -55,6 +58,22 @@ function Playlist(loggerInfo, serializationCtx) {
         inner.sequences = [];
         inner.initialClipIndex = 1;
     }
+    // track when and were playlist was reloaded
```
kalturaguy: @igorshevach  Please remove this code, we can get this info from logs

PR #244, opened 2016-09-01
-1.0-PLAT-PLAT-5976-r

igorshevach:
- when producing playlist.json filter out all 0 duration clips created due to quicks encoder restarts.
- set clip duration to 0 if any of sources contributing to it is empty.


kalturaguy:
@igorshevach  Please add unit-test to cover this use case

PR #247, opened 2016-09-06
-1.0-PLAT-5926

igorshevach:


igorshevach:
scenario  is as follows:
- add ts chunk A = {firstDTS,key_frames_offsets=[0],...} 
- add  ts chunk B = {firstDTS,key_frames_offsets=[0],...} , single gop (single key frame)
- playlist is serialized to disk. chunk B key frame (last gop) is put aside and remembered in last_gop_size field in the playlist.
- entry restart. playlist is serialized from disk. as a result chunk A will be attributed last_gop_size earlier saved in the temp field:
A= {key_frames_offsets=[0,last_gop_size]},
B={key_frames_offsets=[]}
- after playlist window is full decision is taken to pop chunk A out of playlist.
as a result it will get out with the wrong gop along. all of the key frame offsets will get misaligned.

PR #248, opened 2016-09-08
remove network mock

kalturaguy:


lib/NetworkClientFactory.js
```diff
@@ -1,28 +0,0 @@
-/**
```
igorshevach: - will these changes break unit tests?
- did you forget to delete iveDVR/tests/unit/NetworkClientFactory-spec.js as well?
- usages of mockNetwork config var

PR #255, opened 2016-09-15
Insert PlaylistGenerator stop procedure

gadyaari:


lib/entry/LiveEntry.js
```diff
@@ -140,6 +140,10 @@ LiveEntry.prototype.stop = function() {
             // Regardless of the stopping result controller must remove entry from handledEntries list
             return Q.all(flavorsPromises)
                 .then(function() {
+                    that.logger.info("Stopping PlaylistGenerator");
+                    return that.playlistGenerator.stop();
```
igorshevach: what if PlaylistGenerator throws?

PR #258, opened 2016-09-20
implementation, before testing

lilachmaliniak:
seamless failover implementation


lib/entry/crossDCsStreamPersistence.js
```diff
@@ -0,0 +1,208 @@
...
+//   2.2) save playlist. ---> using this.playlistGenerator
+//   2.3) remove from list
+// 3) add sequence to restore _entriesToProcess, when LiveController starts!!!
+
+class CrossDCsStreamPersistence {
+
+    constructor() {
+        let endListPendingEntriesFileName = "endListPending.json";
+        let persistentEntriesMinimalDataFileName = "persistentEntries.json"
+        let localFolder = config.get('localFolderPath') || '.';
```
kalturaguy: localFolderPath?

lib/entry/crossDCsStreamPersistence.js
```diff
@@ -0,0 +1,191 @@
...
+
+//---------------------------------------------------------------
+// Todo: Lilach: PLAT-5023
+// 1) check if any DC provides the entryId:
+//  1.1) in case no DC provides the entry: update the time tag +
+//  1.2) otherwise,
+//      1.2.1) start timer to check if entry is provided
+//      1.2.2) rename playlist.json
+//--------------------------------------------------------------
+CrossDCsStreamPersistence.prototype.streamPersistenceCheck = function (entryId) {
```
kalturaguy: why isn't this written in ES6? (inside the class )?

lib/entry/crossDCsStreamPersistence.js
```diff
@@ -0,0 +1,191 @@
...
+
+    var deferred = Q.defer();
```
kalturaguy: why do you use defer? it's obselete

lib/entry/crossDCsStreamPersistence.js
```diff
@@ -0,0 +1,191 @@
...
+        if (now - this._entriesData[entryId].timestamp >= this.checkStreamIntervalMSec) {
+            this.streamPersistenceCheck(entryId);
+        }
+    });
+
+    return this.flushEntryList(this._entriesToProcess);
+
+};
+
+CrossDCsStreamPersistence.prototype.backwardCompatibleInit = function() {
```
kalturaguy: why do we need this? we don't care about the transition ... it doesn't worth even 5 seconds...

lib/playlistGenerator/PlaylistGenerator.js
```diff
@@ -267,7 +275,8 @@ var doPersist =  function() {
 };
 
 // save playlist
-var persistPlayList = function persistPlayList(force){
+// todo: if stream ended set presentationEndTime to now - playWindow
+var persistPlayList = function persistPlayList(force, ended = false){
```
kalturaguy: it's a bit combersome... i think it'd be better if you expose an interface for the playlist generator that you will setup if it ended or now

lib/playlistGenerator/PlaylistGenerator.js
```diff
@@ -374,10 +384,12 @@ PlaylistGenerator.prototype.stop = function stop (){
     }
     if(isRunning.call(that)){
         that.startPromise = null;
+        that.playlistImp.expirationTime = Data.now;
```
kalturaguy: i'd rather you update the expirationTime every time you update the playlist
so if the process crashes/ machine killed it will automtecly be expired...

lib/playlistGenerator/PlaylistGenerator.js
```diff
@@ -501,7 +513,31 @@ PlaylistGenerator.prototype.reset = function() {
...
+}
 
+// update presentationEndTime and save playlist
+PlaylistGenerator.prototype.persistStopped = function() {
+    return persistPlayList.call(this, true);
+}
 
+// update presentationEndTime and expireTime and save to disk
+PlaylistGenerator.prototype.restoreStopped = function() {
+    this.start();
```
kalturaguy: no need for this , since you will update presentationEndTime and expirationTime when updating the playlist

lib/entry/crossDCsStreamPersistence.js
```diff
@@ -0,0 +1,191 @@
...
+
+    return Q.fcall(function () {
+        if (that._entriesToProcess[entryId]) {
+            logger.info("[%s] entry is already in stream persistence list, ignoring request", entryId);
+            // todo: replace the playlist obj (consult with Guy)
+            // return Q.resolve(); required ???
+            return true;
+        } else {
+            // keep the playlistGenerator obj
+            this._entriesData[entryId] = { playlist: playlistGenerator, timestamp: new Date() };
```
kalturaguy: why not keep the only the playlistgenerator? the timestamp is already kept inside the playlistgenerate, just expose it

lib/entry/crossDCsStreamPersistence.js
lilachmaliniak: backward compatibility... didn't right this code.

lilachmaliniak: after discussion with Guy, the code was removed, as decided.

lib/entry/crossDCsStreamPersistence.js
lilachmaliniak: moved code to ES6 class methods.

lib/entry/crossDCsStreamPersistence.js
lilachmaliniak: removed.

lib/entry/crossDCsStreamPersistence.js
lilachmaliniak: removed

lib/playlistGenerator/PlaylistGenerator.js
lilachmaliniak: code changed. the playlist.json is read and then updated and written to disk directly. Changes in PlaylistGeneration reverted.

lib/playlistGenerator/PlaylistGenerator.js
lilachmaliniak: done.

lib/playlistGenerator/PlaylistGenerator.js
lilachmaliniak: code changed. this method no longer exist.

lib/entry/crossDCsStreamPersistence.js
lilachmaliniak: code updated as mentioned above. PlaylistGenerator obj is not kept anymore. The time kept is the timestamp of last check if entry is provided by any DC.

packager/config/nginx.conf.template
```diff
@@ -49,8 +49,8 @@ http {
         vod_align_segments_to_key_frames on;
 	    vod_response_cache response_cache 128m 30;
 	    vod_max_mapping_response_size 1m;
-        vod_live_mapping_cache live_mapping_cache 128m 2;
-        vod_live_response_cache live_response_cache 128m 2;
+        #vod_live_mapping_cache live_mapping_cache 128m 2;
```
kalturaguy: i think you got this mixed up with other bracnh

lib/playlistGenerator/PlaylistGenerator.js
```diff
@@ -233,9 +233,9 @@ var doPersist =  function() {
     that.emit('beforePlaylistChanged', that);
 
     that.isDirty = false;
-    if(that.playlistImp.inner.liveWindowDuration) {
-        that.playlistImp.inner.presentationEndTime = that.playlistImp.getCurrentTime() + that.playlistImp.inner.liveWindowDuration;
-    }
+    that.playlistImp.inner.presentationEndTime = that.playlistImp.getCurrentTime() + that.playlistImp.inner.liveWindowDuration;
+    that.playlistImp.inner.expirationTime = that.playlistImp.inner.presentationEndTime;
```
kalturaguy: there should be two values:
1. when to return 404
2. when to return end-list

those are different values, in your code it's the same

lib/entry/crossDCsStreamPersistence.js
```diff
@@ -0,0 +1,275 @@
...
+            })
+            .then((isLive) => {
+                entryIsLive = isLive;
+                return qio.read(this._playlistsData[entryId].path);
+            })
+            .then((content) => {
+                let playlist = content;
+                this._playlistsData[entryId].lastProcessedTime = now;
+                if (newEntry) {
+                    logger.debug(`====> [streamPersistenceCheck] [${entryId}] STOPPED STREAMING expiration time set to [${nowStr}]`);
```
kalturaguy: the logger should be consist with all our log (entry id in the begining)
look at the log file

lib/entry/crossDCsStreamPersistence.js
```diff
@@ -0,0 +1,275 @@
...
+
+        logger.debug(`====> [addEntry] adding [${entryId}] to persistence list`);
+
+        if (this._entriesToProcess[entryId]) {
+            logger.debug("====> [addEntry][%s] entry is already in stream persistence list, ignoring request", entryId);
+            return Q.resolve();
+        } else {
+            this._entriesToProcess[entryId] = entryId;
+            this._playlistsData[entryId] = {
+                path: path.join(persistenceFormat.getEntryBasePath(entryId), persistenceFormat.getMasterManifestName()),
```
kalturaguy: why keep the path? entry id is enough, you can always calcualte it

lib/entry/crossDCsStreamPersistence.js
```diff
@@ -0,0 +1,275 @@
...
+
+                    /*         // todo: rethink flash operation location and timing...
+                     return this.flushEntryList(this._entriesToProcess)
+                     .catch((err)=> {
+                     logger.error(`error restoring stopped entries list, error: [${ErrorUtils.error2string(err)}]`);
+                     return Q.resolve(true);
+                     });*/
+                })
+                .catch((err) => {
+                    logger.info(`====> [init] error restoring stopped entries list, error: [${ErrorUtils.error2string(err)}]`);
```
kalturaguy: entry id is missing

lib/entry/crossDCsStreamPersistence.js
```diff
@@ -0,0 +1,275 @@
...
+                });
+        }
+
+    }
+
+    removeEntry(entryId) {
+
+        logger.debug(`====> [removeEntry] removing [${entryId}] from persistence list`);
+        if (this._entriesToProcess.hasOwnProperty(entryId)) {
```
kalturaguy: use .has

lib/entry/crossDCsStreamPersistence.js
```diff
@@ -0,0 +1,275 @@
...
+    addEntry(entryId) {
+
+        logger.debug(`====> [addEntry] adding [${entryId}] to persistence list`);
+
+        if (this._entriesToProcess[entryId]) {
+            logger.debug("====> [addEntry][%s] entry is already in stream persistence list, ignoring request", entryId);
+            return Q.resolve();
+        } else {
+            this._entriesToProcess[entryId] = entryId;
+            this._playlistsData[entryId] = {
```
kalturaguy: why keep extra data structure? it's extra overhead, you just need to keep 
_entriesToProcess the same as we did in 0.15 brach

common/config/config.json.template
```diff
@@ -24,15 +24,45 @@
   "logFileName" : "~/dvr/kLiveController.log",
   "logTimeZoneOffset" : 300,
   "monitorPort": 4390,
+  "persistenceCheckIntervalMillisec": 20000,
+  "stoppedEntriesPath": "~/kLivePersistence",
   "regressionAdapter": {
```
kalturaguy: this doesn't belong to this PR

common/config/config.json.template
```diff
@@ -24,15 +24,45 @@
...
+             }
+        }]
   },
   "diagnostics": {
       "minChunkDurationInSec": 1,
       "maxChunkDurationInSec": 60,
       "errorsWindowSizeInMin" : 5,
       "ptsMaxMisalignmentAllowedInMSec" : 5000
   },
+  "preserveOriginalHLS": {
```
kalturaguy: this doesn't belong to this PR

lib/Controller.js
```diff
@@ -70,8 +72,17 @@ module.exports = (function(){
...
 
+        this.crossDCsStreamPersistence.removeEntry(e.entryId);
+
         return liveEntry.start()
-            .then(function(){
+            .then(function () {
+                return liveEntry;
+            })
+            .then((liveEntry) => {
+                that.crossDCsStreamPersistence.flushEntryList()
```
kalturaguy: why do we call flushEntry list here? the remove Entry that you call few lines ago should do it
the logic of the crossDCsStreamPersistence should not be handled from ourside

lib/Controller.js
```diff
@@ -191,6 +202,9 @@ module.exports = (function(){
             else {
                 logger.info("[%s] Entry is no longer live, stopping it", noLongerLiveEntry);
                 return liveEntry.stop()
+                    .then(() => {
+                        return that.crossDCsStreamPersistence.addEntry(noLongerLiveEntry);
```
kalturaguy: we need to be carfull here, since we don't want exception from addEntry to block our flow,
remove the return

lib/entry/crossDCsStreamPersistence.js
```diff
@@ -0,0 +1,275 @@
...
+
+// todo:
+// consult with Guy whether to keep playlist and just update on disk or read/write...
+// I think its better to keep cache of playlist and just write to disk each update.
+
+class CrossDCsStreamPersistence {
+
+    constructor() {
+
+        this._presentationEndTimeRegex = /("presentationEndTime":\s*)(\d+)/;
```
kalturaguy: this is generic and not per instance should belong ourside the class

lib/entry/crossDCsStreamPersistence.js
```diff
@@ -0,0 +1,275 @@
...
+        // presentationEndTime value becomes past time before it is update by persistence loop.
+        this._timeBudget = 5000; // arbitrary time budget
+    }
+
+    streamPersistenceCheck(entryId, newEntry = false) {
+
+        let entryIsLive = false;
+        let ended = false;
+        let now = Date.now();
+        let nowStr = new Date(now).toLocaleString();
```
kalturaguy: why keep it as string?   dont' use presentationTimeStr and nowStr
also don't use LocaleString

lib/entry/crossDCsStreamPersistence.js
```diff
@@ -0,0 +1,275 @@
...
+            })
+            .then((content) => {
+                let playlist = content;
+                this._playlistsData[entryId].lastProcessedTime = now;
+                if (newEntry) {
+                    logger.debug(`====> [streamPersistenceCheck] [${entryId}] STOPPED STREAMING expiration time set to [${nowStr}]`);
+                    playlist = playlist.replace(this._expireTimeRegex, `$1${now}`);
+                }
+                if (entryIsLive) {
+                    logger.debug(`====> [streamPersistenceCheck] [${entryId}] is LIVE in other DC, presentation end time set to [${presentationTimeStr}]`);
```
kalturaguy: log line already prints the time at the beginning now need for extra string

lib/entry/crossDCsStreamPersistence.js
```diff
@@ -0,0 +1,275 @@
...
+                }
+            });
+
+    }
+
+    flushEntryList() {
+
+        logger.debug(`====> [flushEntryList] updating persistence list on disk`);
+
+        if (Object.keys(this._entriesToProcess).length > 0) {
```
kalturaguy: why all this code?
we can serialize the _entriesToProcess even if they are empty, this will simplify and remove lots of code

lib/entry/crossDCsStreamPersistence.js
```diff
@@ -0,0 +1,275 @@
...
+            let entries2persistStr = JSON.stringify(_.keys(this._entriesToProcess));
```
kalturaguy: please stringfy all  and not just the keys

lib/entry/crossDCsStreamPersistence.js
```diff
@@ -0,0 +1,275 @@
...
+
+            return qio.write(this.stoppedEntriesFileName, entries2persistStr)
+                .then(() => {
+                    logger.debug(`====> [flushEntryList] Stopped entries: ${entries2persistStr}`);
+                })
+                .catch((err) => {
+                    if (err.code === 'ENOENT') {
```
kalturaguy: what is the scenario here?

lib/entry/crossDCsStreamPersistence.js
```diff
@@ -0,0 +1,275 @@
...
+            delete this._entriesToProcess[entryId];
+            delete this._playlistsData[entryId];
+
+            /*        // todo: rethink flash operation location and timing...
+             return this.flushEntryList(this._entriesToProcess)
+             .catch((err) => {
+             logger.error(`failed to get entry server nodes for stopped entry [${entryId}], error: [${ErrorUtils.error2string(err)}]`);
+             return Q.resolve(true);
+             });*/
+            if (Object.keys(this._entriesToProcess).length === 0) {
```
kalturaguy: no need to check that, serialize anyways

lib/entry/crossDCsStreamPersistence.js
```diff
@@ -0,0 +1,275 @@
...
+        }
+
+    }
+
+    removeEntry(entryId) {
+
+        logger.debug(`====> [removeEntry] removing [${entryId}] from persistence list`);
+        if (this._entriesToProcess.hasOwnProperty(entryId)) {
+            delete this._entriesToProcess[entryId];
+            delete this._playlistsData[entryId];
```
kalturaguy: no need for _playlistsData

lib/entry/crossDCsStreamPersistence.js
```diff
@@ -0,0 +1,275 @@
...
+            .then((exists) => {
+                if (exists) {
+                    return qio.read(this.stoppedEntriesFileName);
+                } else {
+                    return Q.reject(false);
+                }
+            })
+            .then((stoppedEntries) => {
+                let pAll = _.map(JSON.parse(stoppedEntries), (entryId) => {
+                    this.addEntry(entryId);
```
kalturaguy: this this is not what you wanted

lib/entry/crossDCsStreamPersistence.js
```diff
@@ -0,0 +1,275 @@
...
+                });
+                return Q.allSettled(pAll);
+            })
+            .catch((err) => {
+                if (_.isObject(err)) {
```
kalturaguy: un-relaistic scneario

lib/entry/crossDCsStreamPersistence.js
```diff
@@ -0,0 +1,275 @@
...
+                    if (err.code != 'ENOENT') {
+                        logger.info(`====> [init] error restoring stopped entries list, error: [${ErrorUtils.error2string(err)}]`);
+                        return Q.reject(err);
+                    } else {
+                        logger.warn(`${this.stoppedEntriesFileName} was not found`);
+                        return Q.resolve();
+                    }
+                } else {
+                    return qio.makeDirectory(stoppedEntriesPath)
```
kalturaguy: better call qio.makeDirectory at the beginning and call finally (regardless if directoy exists or not)

kalturaguy:
@lilachmaliniak  the code review is quite big, I recommend maybe you should look again on 0.15 branch and revisit your code
thanks

packager/config/nginx.conf.template
lilachmaliniak: reverted.

lib/playlistGenerator/PlaylistGenerator.js
lilachmaliniak: yes. the business logic for the 2 values is implemented in crossDCsStreamPersistence.js, per your request the PlaylistGenrator is not involved.

lib/entry/crossDCsStreamPersistence.js
lilachmaliniak: updated.

lib/entry/crossDCsStreamPersistence.js
lilachmaliniak: Each time an entry is added I calculate the path and keep it in memory. It is never saved on disk. I don't see a problem with that.

lib/entry/crossDCsStreamPersistence.js
lilachmaliniak: Init purpose is to restore all stopped entries from file when LiveController starts. Never the less I added log with entry id where relevant.

lib/entry/crossDCsStreamPersistence.js
lilachmaliniak: changed

lib/playlistGenerator/PlaylistGenerator.js
kalturaguy: but you are calculating the same value not seprately, this is the bug

lib/entry/crossDCsStreamPersistence.js
kalturaguy: why do you keep somethign in memory when you can calcualte it every time you need it? please remove redudany data structures

lib/entry/crossDCsStreamPersistence.js
lilachmaliniak: ok

lilachmaliniak: Let's discuss it.

common/config/config.json.template
lilachmaliniak: I did merged from 1.0 on Thursday. this is not my code.

lib/Controller.js
lilachmaliniak: removed.

lib/Controller.js
lilachmaliniak: done.

lib/entry/crossDCsStreamPersistence.js
lilachmaliniak: updated.

lib/entry/crossDCsStreamPersistence.js
lilachmaliniak: removed.

lib/entry/crossDCsStreamPersistence.js
lilachmaliniak: removed.

lib/entry/crossDCsStreamPersistence.js
lilachmaliniak: done.

lib/entry/crossDCsStreamPersistence.js
lilachmaliniak: removed

lib/entry/crossDCsStreamPersistence.js
lilachmaliniak: file not found from some reason.

PR #264, opened 2016-09-28
Diagnostics - fix drift info

gadyaari:


lib/entry/FlavorDownloader.js
```diff
@@ -156,15 +156,14 @@ var FlavorDownloader = (function() {
                         });
                 })
                 .catch(function(err) {
+                    that.logger.error("Error in MP4 conversion: %s, chunkName: %s", ErrorUtils.error2string(err), err.chunkName);
```
kalturaguy: what if err is not an object (string), you will fail in err.chunkName

PR #265, opened 2016-10-03
PLAT 5023 seamless failover with packager

lilachmaliniak:
after code review updates


lib/entry/crossDCsStreamPersistence.js
```diff
@@ -0,0 +1,209 @@
...
+        } else {
+            this._entriesToProcess[entryId] = entryId;
+            logger.info(`[${entryId}] entry added to persistence list`);
+
+            return this.streamPersistenceCheck(entryId, true)
+                .then((ended) => {
+                    logger.info(`[${entryId}] successfully added entry to stream persistence list (ended=${ended}), [playlistPath=${playlistPath.call(this, entryId)}]`);
+                })
+                .then(() => {
+                    this.flushEntryList();
```
kalturaguy: same comment as before, this is not the right this

lib/entry/crossDCsStreamPersistence.js
```diff
@@ -0,0 +1,209 @@
...
+var backendClient = require('../BackendClientFactory.js').getBackendClient();
+var PlaylistGenerator = require('./../playlistGenerator/PlaylistGenerator');
+const endListPendingEntriesFileName = "persistenceList.json";
+const persistenceListPath = config.get('persistenceListPath');
+const kalturaTypes = require('./../kaltura-client-lib/KalturaTypes');
+const presentationEndTimeRegex = /("presentationEndTime":\s*)(\d+)/;
+const expireTimeRegex = /("expirationTime":\s*)(\d+)/;
+
+
+function dateTimeStringFormatter(time) {
```
kalturaguy: please remove this function  use standard building tostring

lib/entry/crossDCsStreamPersistence.js
```diff
@@ -0,0 +1,209 @@
...
+
+        let startTime = Date.now();
+
+        if (startTime - this._lastProcessedTimestampe >= this._persistenceCheckIntervalMillisec) {
+            logger.debug(`Starting to process persistence list`);
+
+            this._lastProcessedTimestampe = startTime;
+
+            let pAll = _.map(this._entriesToProcess, (entryId) => {
+                return this.streamPersistenceCheck(entryId);
```
kalturaguy: this it not the same this.
please read again this and scope in javascript

lib/entry/crossDCsStreamPersistence.js
```diff
@@ -0,0 +1,209 @@
...
+                    playlist = playlist.replace(presentationEndTimeRegex, `$1 ${now}`);
+                }
+                return playlist;
+            })
+            .then((playlist) => {
+                return qio.write(path, playlist);
+            })
+            .then(() => {
+                if (ended) {
+                    return this.removeEntry(entryId, true);
```
kalturaguy: this is no the same this x3...

lib/playlistGenerator/PlaylistGenerator.js
```diff
@@ -233,9 +233,9 @@ var doPersist =  function() {
     that.emit('beforePlaylistChanged', that);
 
     that.isDirty = false;
-    if(that.playlistImp.inner.liveWindowDuration) {
-        that.playlistImp.inner.presentationEndTime = that.playlistImp.getCurrentTime() + that.playlistImp.inner.liveWindowDuration;
-    }
+    that.playlistImp.inner.presentationEndTime = that.playlistImp.getCurrentTime() + that.playlistImp.inner.liveWindowDuration;
```
kalturaguy: expirationTime should be now+ 30 seconds  (when to return 404)
and not same as presentationEndTime

lib/entry/crossDCsStreamPersistence.js
```diff
@@ -0,0 +1,209 @@
...
+
+        let now = Date.now();
+        let presentationEndTime = now + 7200000;
+        let path = playlistPath.call(this, entryId);
+        let ended = !isLive;
+
+        return qio.read(path)
+            .then((content) => {
+                let playlist = content;
+                if (newEntry) {
```
kalturaguy: we don't really need newEntry param, actually you can always change the expire time  to the past?

lib/playlistGenerator/PlaylistGenerator.js
kalturaguy: 1. i think the logic should be in one place, both for this place and the  CrossDCsStreamPersistence
right now it's duplicated 
2. 
if one of the DC is down then
we want the packager to return 404 after X seconds  (in your case it will be the DVR window)

if both DC are down then:
we want to return END-LIST for the DVR window  (now +  liveWindowDuration)

lib/entry/crossDCsStreamPersistence.js
lilachmaliniak: removed function. Using now toString().

lib/entry/crossDCsStreamPersistence.js
lilachmaliniak: removed

lib/playlistGenerator/PlaylistGenerator.js
lilachmaliniak: When the entry is stopped the PlaylistGenerator object is removed. The entry is added to crossDCStreamPersistence and the business logic of stopped and ended is only handled there. The reason there is code for presentationEndTime and expirationTime values in PlaylistGenerator is because you requested that these properties  will always have value. It's by purpose that as long as the entry is live both have future value. As soon as the entry is stopped the crossDCStreamPersistence changes the expiration time to now and this causes the packager to return 404. when entry is not provided by any DC the presentationEndTime is set to now and packager puts END_LIST. I checked that it works.

lib/BackendClient.js
```diff
@@ -118,13 +118,15 @@ var BackendClient = (function(){
         };
         if (item.dvrStatus === kalturaTypes.KalturaDVRStatus.ENABLED) {
 
+            res.dvrEnabled = true;
```
guy1976: ?

common/config/config.json.template
```diff
@@ -24,6 +24,9 @@
   "logFileName" : "~/dvr/kLiveController.log",
   "logTimeZoneOffset" : 300,
   "monitorPort": 4390,
+  "persistenceCheckIntervalSec": 20,
+  "persistenceListPath": "~/kLivePersistence",
```
guy1976: persistenceListFileName

lib/Controller.js
```diff
@@ -191,6 +196,11 @@ module.exports = (function(){
             else {
                 logger.info("[%s] Entry is no longer live, stopping it", noLongerLiveEntry);
                 return liveEntry.stop()
+                    .then(() => {
+                        if (that.handledEntries[noLongerLiveEntry].entryObject.dvrEnabled) {
```
guy1976: call liveEntry.isDvr  (property of liveentry)

lib/entry/LiveEntry.js
```diff
@@ -143,6 +143,7 @@ LiveEntry.prototype.stop = function() {
                 return flavor.stop();
             });
             return Q.all(flavorsPromises)
+            // todo: understand why was the call to playlistGenrator.stop() removed ????
```
guy1976: ?

lib/BackendClient.js
lilachmaliniak: code updated.

lib/Controller.js
lilachmaliniak: done

lib/Controller.js
```diff
@@ -191,6 +196,11 @@ module.exports = (function(){
...
+                            that.crossDCsStreamPersistence.addEntry(noLongerLiveEntry, that.handledEntries[noLongerLiveEntry]);
```
lilachmaliniak: done.

lib/entry/crossDCsStreamPersistence.js
```diff
@@ -0,0 +1,218 @@
...
+
+        let now = Date.now();
+        let path = playlistPath.call(this, entryId);
+        let ended = !isLive;
+
+        return qio.read(path)
+            .then((content) => {
+                let playlist = content;
+
+                if (ended) {
```
kalturaguy: i think it's better to do if (!live) {
   logger.debug(`[${entryId}] ENDED`);
   playlist = playlist.replace(presentationEndTimeRegex, `$1 ${now}`);
  return qio.write(path, playlist).then( ()=> {
                   return this.removeEntry(entryId, true);
  }
}

lib/entry/crossDCsStreamPersistence.js
```diff
@@ -0,0 +1,196 @@
...
+    playlistUpdate(entryId, isLive) {
+
+        let now = Date.now();
+        let path = playlistPath.call(this, entryId);
+
+        return qio.read(path)
+            .then((content) => {
+                let playlist = content;
+
+                if (!isLive) {
```
kalturaguy: add log

lib/BackendClient.js
```diff
@@ -140,6 +142,7 @@ var BackendClient = (function(){
         var filter = new kalturaVO.KalturaLiveStreamEntryFilter();
         filter.isLive = kalturaTypes.KalturaNullableBoolean.TRUE_VALUE;
         filter.hasMediaServerHostname = mediaServerHostname;
+        filter.statusIn = '1,2,3';
```
kalturaguy: please remove

lib/entry/crossDCsStreamPersistence.js
lilachmaliniak: done.

lib/BackendClient.js
lilachmaliniak: removed filter.statusIn

PR #273, opened 2016-10-13
1.0 fix 52bit double precision

igorshevach:
excerpt from https://en.wikipedia.org/wiki/Double-precision_floating-point_format:
IEEE 754 double-precision binary floating-point format:
The IEEE 754 standard specifies a binary64 as having:

Sign bit: 1 bit
Exponent width: 11 bits
Significand precision: 53 bits (52 explicitly stored)

this means that any number > 2^52 (4503599627370496) undergo truncation when converted from 64 bit int.

during ts chunk conversion it is uncommon to come across usage of higher values, hence key frame duration outcome may be inaccurate resulting in mismatch b/w manifest declaration and actual chunk.


lib/playlistGenerator/ConcatSource.js
```diff
@@ -432,8 +442,8 @@ ConcatSource.prototype.checkDiscontinuity = function (fileInfo,track) {
             // calculate immediate value of drift
             that.ptsStats.addSample(correlation);
 
-            // reassign firstDTS in accordance with pts diff to avoid noise
-            if (that.playlist.ptsReference.absolute > 0) {
+            // reassign firstDTS in accordance with pts diff to avoid noise due to jitter
+            if (that.playlist.ptsReference.absolute > -1) {
```
kalturaguy: why not put there that.playlist.ptsReference.absolute!=null?

igorshevach: agreed

PR #283, opened 2016-11-01
1.0 Mock backendclient

kalturaguy:


lib/mocks/BackendClientMock.js
```diff
@@ -84,6 +91,12 @@ var entryServerNodes= {
 module.exports = {
     getLiveEntriesForMediaServer : Q.resolve(entries),
     getEntryInfo : function(entryId) {
+        if (realBackendClient) {
```
lilachmaliniak: live entry statuses are: BROADCASTING, PLAYABLE

PR #291, opened 2016-11-07
name refectory completion _intervalObj -> _handledEntriesTimeout

lilachmaliniak:
From this change on, clips are dropped in one of two situations:
1) clip is empty - no change here
2) max number of clips exceeded - new


lilachmaliniak:
bug was resolved by plat-6253 - liveController can create two entries when machine is loaded

PR #311, opened 2016-11-20
1.1 fix first clip offsets

igorshevach:


common/config/config.json.template
```diff
@@ -26,6 +26,7 @@
   "persistenceCheckIntervalSec": 20,
   "persistenceFileName": "~/dvr/%s_kLivePersistence.json",
   "expirationTimeWindowSec": "30",
+  "maxClipsPerFlavor": "125",
```
kalturaguy: should be 125 not "125"

igorshevach: done

PR #312, opened 2016-11-20
return to update also array of failed updated chunks

ronyad:


lib/recording/RecordingEntrySession.js
```diff
@@ -272,7 +272,8 @@ class RecordingEntrySession {
             }
             return sum;
         },[]);
-        let Mp4FilesUpdatedSorted = _.sortBy(Mp4FilesUpdated, 'startTime')
+        //let Mp4FilesUpdatedSorted = _.sortBy(Mp4FilesUpdated, 'startTime')
+        let Mp4FilesUpdatedSorted = _.shuffle(Mp4FilesUpdated)
```
igorshevach: why shuffle?

lib/recording/RecordingEntrySession.js
```diff
@@ -283,16 +284,22 @@ class RecordingEntrySession {
...
-                else {
-                    this.logger.debug("LinkAndUpdateJson, Successfully update: %s", chunksNameChained)
-                }
+                this.logger.debug("About to parse result of chunks: %s", chunksNameChained)
+                this.parseUpdateJsonResult(results[1].value)
                 return Q.resolve(chunksNameChained)
             })
     }
 
+    parseUpdateJsonResult(result){
```
igorshevach: what are you trying to achieve?
in 1.2 there is code change: instead of returning array of removed chunks from update an event is raised.
so basically this code is obsolete.

PR #314, opened 2016-11-21
1.2 filter all kaltura id3 tag object types save kaltura syncpoint

igorshevach:


node_addons/FormatConverter/include/Utils.h
```diff
@@ -160,94 +160,28 @@ return "";\
...
-            av_log(NULL, AV_LOG_DEBUG,"extractUnixTimeMsecFromId3Tag. error %s code: %d => %s", ex.what(), ex.code(), reason);
-            throw;
-            
-        }
-#else
-        string pattern ("\"timestamp\":");
-        size_t n = s.find(pattern);
+        // only sync points of type KalturaSyncPoint are accepted...
+        string type("\"objectType\":\"KalturaSyncPoint\"");
+        size_t n = s.find(type);
```
kalturaguy: won't it better todo regex? to parse the entire string?

igorshevach: c++11 does not implement regex spec well on centos

PR #319, opened 2016-11-24
Fix flavors missing

kalturaguy:


kalturaguy:
@lilachmaliniak  Please review

PR #320, opened 2016-11-24
Create KS with disableEntitlment to avoid end user entitlment

yossipapi:


lib/BackendClient.js
```diff
@@ -73,7 +73,7 @@ var BackendClient = (function(){
                     client.setSessionId(results);
                     deferred.resolve();
                 }
-            }, adminSecret, null, kalturaTypes.KalturaSessionType.ADMIN, partnerId, null, null);
+            }, adminSecret, null, kalturaTypes.KalturaSessionType.ADMIN, partnerId, null, "all:*,disableentitlement");
```
kalturaguy: i think it's better to add this  into the config.template 
"backendClient" : {
    "adminSecret" :"admin secret",
    "partnerId" : 12345,
    "serviceUrl" : "http://serviceURL",
    "ksSessionRefreshIntervalMinutes" : 720,
    "ksPrivileges":  "all:*,disableentitlement"
  },

yossipapi:
@kalturaguy updated with your suggestion please review.

kalturaguy:
@yossipapi  Looks OK, but you need to update OLD liveDVR saas tag to match the new config

yossipapi:
I know ... I see that the latest tag was released from master so should I do my changes on teh master branch?

PR #322, opened 2016-11-27
1.2 plat 5838 live consistency tests

lilachmaliniak:


common/PersistenceFormat.js
```diff
@@ -61,52 +67,151 @@ class PersistenceFormatBase {
...
-            var matched = tsChunktMatch.exec(tsChunkName);
-            if (matched) {
-                return matched[1] + '-' + matched[2] + '.mp4';
+if (!preserveOriginalHLS.enable) {
+    module.exports = new PersistenceFormatBase();
+} else {
+
+    function getEntryPathHelper(entryPath) {
+
+        let basePath = `${rootFolder}/${entryPath}-`;
```
igorshevach: please use path.join whenever you build paths

common/PersistenceFormat.js
```diff
@@ -61,52 +67,151 @@ class PersistenceFormatBase {
...
+
+
+}
+
 
-        compressChunkName(tsChunkName) {
-            var matched = tsChunktMatch.exec(tsChunkName);
-            if (matched) {
-                return matched[1] + '-' + matched[2] + '.mp4';
+if (!preserveOriginalHLS.enable) {
```
igorshevach: it looks like checking for dirHierarchy && preserveOriginalHLS.enable will simplify the code

common/PersistenceFormat.js
lilachmaliniak: done in all relevant places.

common/PersistenceFormat.js
lilachmaliniak: changed logic the base class will be created either if hls preserve is disabled or if using hierarchy of files. Renamed the class used when hls is perserved without hierarchy from preserveOriginalHLSFormat to SimplfiedPersistenceFormat.

lilachmaliniak:
will be merged to 1.3.

PR #323, opened 2016-11-27
1.2 PLAT 6121 Limit max clips per flavor

lilachmaliniak:


lilachmaliniak:
all relevant changes merged already from another branch.

PR #329, opened 2016-11-30
1.1 PLAT 6121 Limit max clips per flavor

lilachmaliniak:


lilachmaliniak:
will be merged to 1.3.

PR #334, opened 2016-12-01
1.1 PLAT 6121 Limit max clips per flavor

lilachmaliniak:
The fix is already merged into 1.2 with changes to support DASH.
This pull request mainly adds log messages.


lilachmaliniak:
No change here is significant, all can be skipped.

PR #338, opened 2016-12-01
bug

ronyad:


lib/recording/RecordingEntrySession.js
```diff
@@ -231,7 +231,7 @@ class RecordingEntrySession {
             let chunkName = path.basename(Mp4File.path)
             if (flavorId == undefined){
                 let arrayPath = Mp4File.path.split(path.sep);
-                var flavorId = arrayPath[arrayPath.length - 3];
+                var flavorId = arrayPath[arrayPath.length - 2];
```
kalturaguy: suggest you will use PersistanceFormat

PR #343, opened 2016-12-05
Fixing init.d script

GElkayam:


GElkayam:
Added Folder creation on start

PR #345, opened 2016-12-06
preserve hls exactly as entry stream is saved in production

lilachmaliniak:
After this commit every preservation shall override previous unless
either copied/renamed or suing SimulateStreams + path parameter (with
unique value).

### DIFF
--- a/common/PersistenceFormat.js
+++ b/common/PersistenceFormat.js
@@ -9,7 +9,6 @@ var _ = require('underscore');
 var Q = require('q');
 
 const tsChunktMatch =  new RegExp(/media-([^_]+).*?([\d]+)\.ts.*/);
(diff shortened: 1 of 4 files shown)


common/PersistenceFormat.js
```diff
@@ -9,7 +9,6 @@ var _ = require('underscore');
 var Q = require('q');
 
 const tsChunktMatch =  new RegExp(/media-([^_]+).*?([\d]+)\.ts.*/);
-var preserveOriginalHLS = config.get('preserveOriginalHLS').enable;
 const rootFolder = config.get('rootFolderPath');
 
 class PersistenceFormatBase {
```
igorshevach: change name to PersistenceFormat since 'Base' isn't relevant anymore

lilachmaliniak: done. Class name is now PersistenceFormat.

PR #346, opened 2016-12-06
reduced segment duration to 4secs + window to 40secs

lilachmaliniak:


lilachmaliniak:
current changes are not the way code should support low latency.

PR #351, opened 2016-12-07
1.3 PLAT 6522 persistence list restore fail

lilachmaliniak:
plat-6522 fixed.

### DIFF
--- a/lib/entry/crossDCsStreamPersistence.js
+++ b/lib/entry/crossDCsStreamPersistence.js
@@ -16,6 +16,7 @@ const presentationEndTimeRegex = /("presentationEndTime":\s*)(\d+)/;
 const util = require('util');
 const fsUtils = require('./../utils/fs-utils');
 
+const entryPattern = new RegExp(/^[0,1]_[\w]+/);
 
 function playlistPath(entryId) {
     return path.join(persistenceFormat.getEntryBasePath(entryId), persistenceFormat.getMasterManifestName());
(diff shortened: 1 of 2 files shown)


lib/entry/crossDCsStreamPersistence.js
```diff
@@ -146,26 +148,33 @@ class CrossDCsStreamPersistence {
                 }
             })
             .then((stoppedEntries) => {
-                let pAll = _.map(JSON.parse(stoppedEntries), (entryId) => {
-                    logger.debug(`[${entryId}] STOPPED ENTRY is being restored from ${this._persistenceFileName}`);
-                    this.addEntry(entryId);
+                let pAll = _.map(stoppedEntries.split(','), (entryId) => {
+                    logger.debug(`Adding [${entryId}], STOPPED ENTRY, from ${this._persistenceFileName}, live entries persistence list`);
+                    this._entriesToProcess[entryId] = entryId;
```
kalturaguy: why aren't you calling addEntry?

lib/entry/crossDCsStreamPersistence.js
```diff
@@ -146,26 +148,33 @@ class CrossDCsStreamPersistence {
...
     }
 
     flushEntryList() {
-        let persistenceListStr = JSON.stringify(this._entriesToProcess);
-        logger.debug(`Updating persistence list on disk, content: [${persistenceListStr}], save in: [${this._persistenceFileName}]`);
 
-        return qio.write(this._persistenceFileName, persistenceListStr)
+        let persistenceListStr = this.toJSON();
+
+        return fsUtils.writeFileAtomically(this._persistenceFileName, persistenceListStr)
```
kalturaguy: revert back the stringify

lib/entry/crossDCsStreamPersistence.js
```diff
@@ -200,7 +209,7 @@ class CrossDCsStreamPersistence {
     }
 
     toJSON() {
-        return Object.keys(this._entriesToProcess);
+        return Object.keys(this._entriesToProcess).toString();
```
kalturaguy: not good... since this is what is returned to the diagnostics, and we don't want it as string but as object?

lib/entry/crossDCsStreamPersistence.js
lilachmaliniak: log moved into the then and catch clause.

lib/entry/crossDCsStreamPersistence.js
lilachmaliniak: Reverted.

lilachmaliniak:
I do. line 160.

PR #352, opened 2016-12-07
1.3 plat 6522 persistence list restore fail

lilachmaliniak:


lib/entry/crossDCsStreamPersistence.js
```diff
@@ -76,7 +78,15 @@ class CrossDCsStreamPersistence {
                 logger.debug(`[${entryId}] is already in stream persistence list, ignoring request`);
                 return Q.resolve();
             } else {
-                this._entriesToProcess[entryId] = entryId;
+                // verify that entry id is valid
```
kalturaguy: why do we need to check this?

lib/entry/crossDCsStreamPersistence.js
```diff
@@ -149,30 +149,31 @@ class CrossDCsStreamPersistence {
                 let pAll = _.map(JSON.parse(stoppedEntries), (entryId) => {
                     logger.debug(`[${entryId}] STOPPED ENTRY is being restored from ${this._persistenceFileName}`);
                     this.addEntry(entryId);
-                });
-                return Q.allSettled(pAll);
+                    });
```
kalturaguy: fix indeitation

PR #361, opened 2016-12-14
1.3-PLAT-6617

igorshevach:
- determining baseClipTime is deferred until all flavors are ready
 - playlist is not written until first clip duration > 0

### DIFF
--- a/lib/playlistGenerator/Playlist.js
+++ b/lib/playlistGenerator/Playlist.js
@@ -402,7 +402,7 @@ var calcClipStartTimeAndDuration = function(index){
         that.inner.durations[index].value = retVal.max - that.inner.clipTimes[index];
 
         // special case for first clip
-        if (index === 0) {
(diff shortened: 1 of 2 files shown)


lib/playlistGenerator/Playlist.js
```diff
@@ -437,19 +437,18 @@ Playlist.prototype.recalculateOffsetsAndDuration = function recalculateOffsetsAn
...
     });
 
-    if(minMaxInfos.length) {
+    if(minMaxInfos.length && _.every(minMaxInfos,r=>r.valid)) {
         retVal.min = minMaxInfos.first.min;
         retVal.max = minMaxInfos.last.max;
-        that.updateGap();
     }
 
+    that.updateGap();
```
lilachmaliniak: check  that taking the call to updateGap from the condition cannot cause any new bug.

PR #373, opened 2017-01-04
1.4 plat 6707

ronyad:


lib/entry/LiveEntry.js
```diff
@@ -57,7 +62,16 @@ function onNewTsFilesDownloaded(flavorObj, streamDurationAdditionInSec) {
         }
     }
 }
-
+function triggerEvent(event, ...params) {
+    let promises = []
+    _.each(this.observers, (observer)=>{
+        if (observer.hasOwnProperty(event)) {
+            let promise = observer[event](...params);
```
kalturaguy: add try catch...

lib/BackendClient.js
```diff
@@ -118,7 +118,6 @@ var BackendClient = (function(){
             "recordStatus" : item.recordStatus
         };
         if (item.recordStatus !== kalturaTypes.KalturaRecordStatus.DISABLED){
-            res.recordingSessionDuration = config.get('recording').recordingSessionDurationInSec * 1000;
```
kalturaguy: do you use it somewher?

lib/Controller.js
```diff
@@ -20,7 +20,7 @@ var util = require('util');
 var MonitorServer=require('./MonitorServer.js');
 var CrossDCsStreamPersistence = require('./entry/crossDCsStreamPersistence');
 var backendClient=require('./BackendClientFactory.js').getBackendClient();
-var RecordingManager = require('./recording/RecordingManager');
+var RecordingManager = require('./plugins/recording/RecordingManager');
```
kalturaguy: maybe we need some plugin initiator here?

lib/entry/LiveEntry.js
```diff
@@ -33,16 +34,20 @@ function LiveEntry(entryObject) {
...
 }
 util.inherits(LiveEntry, events.EventEmitter);
 
-function onNewTsFilesDownloaded(flavorObj, streamDurationAdditionInSec) {
+function onNewTsFilesDownloaded(flavorObj, streamDurationAdditionInSec, newMp4Files) {
     // Once entry has the configured minChunksForPlayback call registerMediaServer->Playing every time chunklist is
     // updated. On the first time change its state and raise isLive for player, from there on update database
     // that entry is still alive every 60 seconds.
-    var that = this;
+        try{
```
kalturaguy: try catch is not need here, move it inside the triggerEvent

lib/entry/LiveEntry.js
kalturaguy: what if observer[event] doesn't return promise?  does it work?

lib/plugins/recording/RecordingManager.js
```diff
@@ -4,21 +4,21 @@
...
-const logger = require('../../common/logger').getLogger("RecordingManager");
-const configRecording = config.get("recording")
-const kalturaTypes = require('../kaltura-client-lib/KalturaTypes');
+const logger = require('../../../common/logger').getLogger("RecordingManager");
+const kalturaTypes = require('../../kaltura-client-lib/KalturaTypes');
 const unknownEntry = "UNKNOWN"
-const backendClient = require('../BackendClientFactory.js').getBackendClient();
-class RecordingManager {
+const backendClient = require('../../BackendClientFactory.js').getBackendClient();
+const basePlugin = require('../BasePlugin')
```
kalturaguy: rename to BasePlugin

lib/BackendClient.js
ronyad: Yes, but I move it to recordingEntrySession.

lib/Controller.js
ronyad: ?

PR #433, opened 2017-03-27
PLAT-7192: set keepOriginalFile to force mv and cp which causes setRe…

yossipapi:
…cordedContnet request timeouts + update clients


yossipapi:
@kalturaguy please review

PR #436, opened 2017-04-25
Support updating audio language for recorded playlist

yossipapi:


lib/entry/FlavorDownloader.js
```diff
@@ -388,8 +388,18 @@ var FlavorDownloader = (function() {
     }
 
     FlavorDownloader.prototype.setLanguage = function(language) {
+        var that = this;
         this.language = language;
         this.playlistGenerator.setSequenceLanguage(this.flavor, language);
+
```
kalturaguy: not a good location... the handling of recording should all be done in the RecordingEntrySession.js

@lilachmaliniak   Please move this into RecordingEntrySession where we create the playlistGenerator and have access to flavorsObjArray

PR #441, opened 2017-05-11
correctly support extract of flavor id when recording has single flavor

lilachmaliniak:
plat-7373


kalturaguy:
@lilachmaliniak  OK

PR #458, opened 2017-05-28
merge LiveRecoder update (plat-7473)

lilachmaliniak:
now checking if queue is full every ready file (job).


liveRecorder/Tasks/TaskRunner.py
```diff
@@ -94,12 +94,12 @@ def schedule_job(self):
 
     def move_and_add_to_queue(self, src_dir):
```
kalturaguy: Maybe we need to leave this also?

PR #462, opened 2017-06-08
PLAT-7525: Fix scenario when server would return a new recordedEntryI…

gadyaari:
…d and recordingManager would ignore it.

### DIFF
--- a/lib/recording/RecordingEntrySession.js
+++ b/lib/recording/RecordingEntrySession.js
@@ -74,10 +74,10 @@ class RecordingEntrySession {
         this.recordingSessionId = entryId + '-' + recordedEntryId;
         this.logger = loggerBase.getLogger('RecordingEntrySession', '[' + this.recordingSessionId + '] ');
         this.logger.info("Starting a new recording session");
-        this.serializedActionsPromise = Q.resolve()
+        this.serializedActionsPromise = Q.resolve();
         this.getDuration = ()=>{
             return this.playlistGenerator.getTotalDuration()
-        }
+        };
         this.serverType = entryObj.serverType
 
     }
@@ -121,18 +121,18 @@ class RecordingEntrySession {
      In order to check that all chunks in live directory exist in recording, we check first that all chunks in recording are in json
      */
     updateEntryDuration(force){
-        let duration = this.getDuration()
-        if (force || duration>this.lastUpdateDuration + recordingUpdateEntryTimeIntervalMs) {
-            this.lastUpdateDuration = duration
-            this.logger.info("New last update duration %s", duration)
-            return backendClient.updateEntryDuration(this.entryId, duration).then(()=> {
-                this.logger.info("Successfully update liveEntry %s with duration %s", this.entryId, duration)
-            })
+        let duration = this.getDuration();
+        if (force || duration > this.lastUpdateDuration + recordingUpdateEntryTimeIntervalMs) {
+            this.lastUpdateDuration = duration;
+            this.logger.info(`[${this.entryId}] New last update duration ${duration}`);
+            return backendClient.updateEntryDuration(this.entryId, duration).then((result)=> {
+                    this.logger.info(`[${this.entryId}] Successfully updated liveEntry with duration ${duration}`)
+                })
         }
-        return Q.resolve()
+        return Q.resolve();
     }
-    restoreSession() {
 
+    restoreSession() {
         return this.accessibleRecording() //in order to check if file exist, first need to load Json form disk
             .then(()=>{
                 return this.checkFilesFromRecordingFolder()
@@ -147,6 +147,7 @@ class RecordingEntrySession {
                 this.logger.error("Failed to restore session : %s", ErrorUtils.error2string(err))
             })
     }
+
     checkMissingChunks  (downloadedChunks) {
         let missingChunks = [];
         _.each(downloadedChunks, (filePath)=> {
@@ -192,7 +193,6 @@ class RecordingEntrySession {
     }
 
     findMissingChunksAndGetMetadata(directoriesToCheck){
-
         return pGlob(directoriesToCheck)
             .then( (downloadedChunks) => {
                 let missingChunks = this.checkMissingChunks(downloadedChunks);
@@ -211,7 +211,6 @@ class RecordingEntrySession {
      This function is parsing the metadata of all files
      */
     getAllMetaData(missingChunks) {
-
         function getMetaDataChunk(filePath) {
             this.logger.debug("About to parse meta data for the file %s", filePath)
             return qio.stat(filePath)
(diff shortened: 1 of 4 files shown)


lib/recording/RecordingManager.js
```diff
@@ -291,53 +304,59 @@ class RecordingManager {
...
-                logger.warn("[%s-%s][stopRecording] Entry duration hasn't changed, do nothing. start [%s] current[%s], totalAddedChunks [%s]", entryId, recordedEntryId, recordingEntrySession.getDuration(), recordingEntrySession.startDuration , totalAddedChunks)
-                delete this.recordingList[entryId];
-                return qio.write(doneFilePath, 'done')
+    stopRecording(entryId) {
+        if (this.recordingList[entryId]) {
+            let recordingEntrySession = this.recordingList[entryId];
+            let doneFilePath = path.join(recordingEntrySession.recordingSessionPath, 'done')
+            if (!recordingEntrySession) {
+                logger.warn("[%s][stopRecording] Can't stop recording, entry doesn't exist on recording list", entryId);
+                return
```
lilachmaliniak: return promise

lib/recording/RecordingManager.js
```diff
@@ -43,119 +44,136 @@ class RecordingManager {
...
                                 }
                             }
                         })
                     })
             }
-            const getRecordingEntrySession = ()=>{
-                let recordingEntrySession = null
+
+            function getRecordingEntrySession(entryObj) {
                 let entryId = entryObj.entryId;
```
kalturaguy: you move this to be a function, so every this inside is a different scope... so proably everything is broken...

lib/recording/RecordingManager.js
```diff
@@ -291,53 +304,59 @@ class RecordingManager {
...
-            let totalAddedChunks = _.reduce(recordingEntrySession.flavorsChunksCounter,(memo,flavorChunks)=>{
-                return memo+ flavorChunks
-            }, 0)
-            //todo should also print warning if # of chunks is not equall between all flavors!
-            let recordingSessionDuration = recordingEntrySession.getDuration() - recordingEntrySession.startDuration
-            if (! recordingSessionDuration > 0 ){
-                logger.warn("[%s-%s][stopRecording] Entry duration hasn't changed, do nothing. start [%s] current[%s], totalAddedChunks [%s]", entryId, recordedEntryId, recordingEntrySession.getDuration(), recordingEntrySession.startDuration , totalAddedChunks)
-                delete this.recordingList[entryId];
-                return qio.write(doneFilePath, 'done')
+    stopRecording(entryId) {
```
lilachmaliniak: General comment:
stopRecording return promise that caller function doesn't handle. Please update code and test.

lib/recording/RecordingManager.js
```diff
@@ -43,119 +44,136 @@ class RecordingManager {
...
+                    });
+                })
+                .catch((err)=> {
+                    // In case of error, pass message on and print to log
+                    return Q.reject(err);
+                });
         }).catch((err)=>{
-            logger.error("[%s-%s] Error while try to start recording : %s, delete object", entryObj.entryId, entryObj.recordedEntryId, ErrorUtils.error2string(err))
+            logger.error(`[${entryObj.entryId}-${entryObj.recordedEntryId}] Error while try to start recording : ${ErrorUtils.error2string(err)}, delete object`);
             delete this.recordingList[entryObj.entryId]
```
lilachmaliniak: check if there is item in recording list before delete. Please test this flow.

lib/recording/RecordingManager.js
```diff
@@ -43,119 +44,136 @@ class RecordingManager {
...
 
-        setInterval(()=>{
+        setInterval(()=> {
             this.handleRecordingEntries()
-        }, recordingTimeInterval, this)
+        }, recordingTimeInterval, this);
     }
 
     startRecording(entryObj, flavorsObjArray){
         return Q.resolve().then(()=> {
```
lilachmaliniak: General comment: if startRecording returned promise is rejected live entry will fail to start. I don't think we want that but is for product to decide. Please consult with Guy & Yaron and test the reject flow as well.

lib/recording/RecordingManager.js
gadyaari: I pass the function entryObj as a parameter, which is everything needed for it to run.
Also, tested it and nothing got broken. If still want me to change it be let me know

lib/recording/RecordingManager.js
gadyaari: If property doesn't exist in object delete has no effect and returns, so don't see the need to add another checkup here.
You can see the following: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/delete

lib/recording/RecordingManager.js
gadyaari: Don't really understand what do you mean

lib/recording/RecordingManager.js
gadyaari: done.

lib/recording/RecordingManager.js
gadyaari: Good point. Will do

lib/recording/RecordingManager.js
```diff
@@ -39,123 +40,141 @@ class RecordingManager {
...
                         })
                     })
             }
-            const getRecordingEntrySession = ()=>{
-                let recordingEntrySession = null
+
+            function getRecordingEntrySession(entryObj) {
                 let entryId = entryObj.entryId;
-                if (!("recordedEntryId" in entryObj && entryObj.recordedEntryId != undefined)){
+                if (!("recordedEntryId" in entryObj && entryObj.recordedEntryId != undefined)) {
```
kalturaguy: replace entryObj.recordedEntryId != undefined with entryObj.recordedEntryId

lib/recording/RecordingEntrySession.js
```diff
@@ -121,18 +121,18 @@ class RecordingEntrySession {
...
-            this.logger.info("New last update duration %s", duration)
-            return backendClient.updateEntryDuration(this.entryId, duration).then(()=> {
-                this.logger.info("Successfully update liveEntry %s with duration %s", this.entryId, duration)
-            })
+        let duration = this.getDuration();
+        if (force || duration > this.lastUpdateDuration + recordingUpdateEntryTimeIntervalMs) {
+            this.lastUpdateDuration = duration;
+            this.logger.info(`[${this.entryId}] New last update duration ${duration}`);
+            return backendClient.updateEntryDuration(this.entryId, duration).then((result)=> {
+                    this.logger.info(`[${this.entryId}] Successfully updated liveEntry with duration ${duration}`)
```
lilachmaliniak: Gad, if you return result then also log it.

PR #465, opened 2017-06-12
handle mismatch between provisioned stream and live entry transcode profile

lilachmaliniak:


liveRecorder/Tasks/UploadTask.py
```diff
@@ -86,17 +86,23 @@ def check_replacement_status(self, partner_id):
...
+        else: # upload with flavor id to indicate that transcoding was already done
+            self.backend_client.set_recorded_content_local(partner_id, self.entry_id, file_full_path,
                                                        str(float(self.duration)/1000), self.recorded_id, flavor_id)
 
     def run(self):
         try:
             mode = get_config('mode')
             is_first_flavor = True
+            num_flavors = len(self).mp4_files_list
+
```
kalturaguy: there could be a valid scenario where user is streaming only 1 flavor...

PR #466, opened 2017-06-13
PLAT-7225: Support 24/7 recordings

gadyaari:


lib/entry/LiveEntry.js
```diff
@@ -73,14 +72,14 @@ function onFlavorDownloaderStopped(flavor) {
 
 function updateFlavorsLanguage() {
     // Iterate through the flavors array and erase duplicate flavors if they exists
-    _.uniq(this.flavorsObjArray);
-    let flavors = _.map(this.flavorsObjArray, (f)=> {
+    _.uniq(this.entryObject.flavorsObjArray);
```
kalturaguy: redudant

lib/recording/RecordingManager.js
```diff
@@ -182,6 +182,27 @@ class RecordingManager {
...
 
+    splitRecording(entryId) {
+        let recordingSession = this.recordingList[entryId];
+        if (!recordingSession.mp4FilesQueueLock) {
+            recordingSession.mp4FilesQueueLock = true;
+            logger.info(`[${entryId}] Recording entry ${this.recordingList[entryId].recordedEntryId} has passed the limit of recording duration - ${this.recordingMaxDurationInMs/1000} seconds. Split session`);
+            return recordingSession.reportServerMaxRecordingDuration()
+                .then((newRecordedEntryId)=> {
+                    logger.info(`[${entryId}] Successfully created new recordedEntryId: [${newRecordedEntryId}]; restart session`);
+                    recordingSession.entryObject.recordedEntryId = newRecordedEntryId;
```
lilachmaliniak: I think you just overrun the recordedEntryId of the recrdingEntrySession in the recordingList. I suggest that the value returned from BE client will be entryObject instead of the recorded entry Id. then pass to the call to startRecording the new EntryObject.
check additional alternative to solve here:
https://stackoverflow.com/questions/10510063/javascript-object-reference-linked-to-object-in-array
using Object.assign() is one option.

PR #468, opened 2017-06-27
fix liveController processes status check

lilachmaliniak:


serviceWrappers/linux/kLiveController
```diff
@@ -93,22 +97,18 @@ restart() {
...
-            else
-            echo "${NAME}_${prefix} is not running but there is stale pidFile: $PIDFILE"
-            RETVAL=$(( $RETVAL + 1 ))
- 	    fi
-        else
-            echo "${NAME}_${prefix} is not running."
-	        RETVAL=$(( $RETVAL + 1 ))
-        fi
+            status -p ${PIDFILE}
+            if [ ! $? -eq 0 ]; then
```
jessp01:
```
if [ ! $? -eq 0 ]; then
```
should be:
```
if [ $? -ne 0 ]; then
```
Also, as I wrote in https://kaltura.atlassian.net/browse/PLAT-7392
However, there's another problem here: you're checking the status of two different PIDs in the same function [in a for loop] but you do not return in the event status -p /var/run/kLiveController_0.pid /opt/kaltura/liveController/latest/lib/App.js fails, as a result, there may be a case where /var/run/kLiveController_0.pid does not contain a valid PID but /var/run/kLiveController_1.pid does and the status action will therefore return with RC 0, even though one of the two procs [the first one] is not running.
In the event /var/run/kLiveController_0.pid does not contain a valid PID, you should return with $?.

serviceWrappers/linux/kLiveController
```diff
@@ -93,28 +105,28 @@ restart() {
...
-	        RETVAL=$(( $RETVAL + 1 ))
+        PIDFILE="${PID_DIR}/${NAME}_${prefix}.pid"
+        if [ -r ${PIDFILE} ]; then
+            status -p ${PIDFILE}
+	        if [ $? -ne 0 ]; then
+	            RETVAL=$?
+	            return ${RETVAL}
+	        fi
+	    else
+	        echo "${PIDFILE} is empty, ${NAME} is not running (verify by: ps aux | grep node)"
```
jessp01: need to properly indent, also, the message is wrong. This point in the code will only be reached if PIDFILE is not readable, not if it's empty.

serviceWrappers/linux/kLiveController
```diff
@@ -32,18 +37,18 @@ HOME=${HOME:-~}
 
 loadNvm() {
     # This loads nvm
-    cd $KLIVE_CONTROLLER_PREFIX
-    [ -z "$NVM_DIR" ] && NVM_DIR="$HOME/.nvm"
-    if [ -s "$NVM_DIR/nvm.sh" ] ; then
-        source "$NVM_DIR/nvm.sh"
+    cd ${KLIVE_CONTROLLER_PREFIX}
```
jessp01: Why is cd needed here? if you just need to check that ${KLIVE_CONTROLLER_PREFIX}/.nvmrc isn't empty then use the full path and check.

serviceWrappers/linux/kLiveController
```diff
@@ -53,20 +58,20 @@ start() {
     loadNvm
     RETVAL=0
     for prefix in 0 1 ; do
-        PIDFILE="$PID_DIR/${NAME}_${prefix}.pid"
-        LOGFILE="$LOG_DIR/${NAME}_${prefix}-forever.log"
-        if [ -f $PIDFILE ]; then
+        PIDFILE="${PID_DIR}/${NAME}_${prefix}.pid"
+        LOGFILE="${LOG_DIR}/${NAME}_${prefix}-forever.log"
+        if [ -f ${PIDFILE} ]; then
             echo "${NAME}_${prefix} already running"
```
jessp01: call status. Just because there is a PID file does not mean we are running.

serviceWrappers/linux/kLiveController
lilachmaliniak: changed

serviceWrappers/linux/kLiveController
lilachmaliniak: corrected

serviceWrappers/linux/kLiveController
lilachmaliniak: updated echo message

lilachmaliniak:
updated kLiveController and liveRecorder scripts
highlights:
- use functions script to get process status
- call . /opt/rh/python27/enable to add python path to PATH 
- removed kill (called from stop if needed)
usage kLiveController {start|stop|restart|status}

liveRecorder/serviceWrappers/linux/liveRecorder
```diff
@@ -18,22 +18,29 @@
...
 APPDIR=${SCRIPT_DIR%/serviceWrappers/linux}
 NAME=${APP_NAME}
 APPLICATION_PATH=${APPDIR}/main.py
 LOG_DIR="/var/log/${APP_NAME}"
 PID_DIR="/var/run"
-PIDFILE="$PID_DIR/${NAME}.pid"
-LOGFILE="$LOG_DIR/${NAME}-service.log"
+PIDFILE="${PID_DIR}/${NAME}.pid"
+LOGFILE="${LOG_DIR}/${NAME}-service.log"
+LOGFILE="${LOG_DIR}/${NAME}-service.log"
```
kalturaguy: duplicate line

liveRecorder/serviceWrappers/linux/liveRecorder
```diff
@@ -44,30 +51,41 @@ createFolders() {
...
 		else
-			echo "starting ${NAME} failed. please see $LOGFILE "
+			echo "starting ${NAME} failed. please see ${LOGFILE}"
 		fi
 	fi
 }
 
 stop() {
-    if [ -f $PIDFILE ]; then
+    if [ -f ${PIDFILE} ]; then
```
jessp01: Again, you need to check that the PID is really alive, the fact the PID file is present is not enough.
Call status.

serviceWrappers/linux/kLiveController
```diff
@@ -32,18 +37,17 @@ HOME=${HOME:-~}
...
-        source "$NVM_DIR/nvm.sh"
-        if [ -s .nvmrc ] ; then
+    [ -z "${NVM_DIR}" ] && NVM_DIR="${HOME}/.nvm"
+    if [ -s "${NVM_DIR}/nvm.sh" ] ; then
+        source "${NVM_DIR}/nvm.sh"
+        if [ -s ${KLIVE_CONTROLLER_PREFIX}/.nvmrc ] ; then
             nvm use || nvm install
         else
-            echo "no .nvmrc in $PWD , Exiting"
+            echo "no .nvmrc in ${PWD} , Exiting"
```
jessp01: Change this to no " ${KLIVE_CONTROLLER_PREFIX}/.nvmr", CWD should never be relied upon.

serviceWrappers/linux/kLiveController
```diff
@@ -79,11 +84,18 @@ stop() {
         if [ -f $PIDFILE ]; then
```
jessp01: Same story, use status to verify the PID in PIDFILE is a valid one.

liveRecorder/serviceWrappers/linux/liveRecorder
lilachmaliniak: done

serviceWrappers/linux/kLiveController
lilachmaliniak: done

liveRecorder/serviceWrappers/linux/liveRecorder
lilachmaliniak: done

serviceWrappers/linux/kLiveController
lilachmaliniak: done

PR #469, opened 2017-06-27
updated build_ffmpeg.sh

lilachmaliniak:


lilachmaliniak:
corrected the paths used for download and extract ffmpeg

PR #470, opened 2017-06-27
build scripts update

lilachmaliniak:


build_scripts/build_ffmpeg.sh
```diff
@@ -22,68 +22,43 @@ fi
...
-    [ ${OS} == "Linux" ] && confCmd="${confCmd} --enable-pic"
+echo "configuring ffmpeg..."
+eval "${confCmd}"
 
-      echo "configuring ffmpeg..."
-      eval "${confCmd}"
+echo "version=${FFMPEG_VERSION} ${confCmd}" > ${configFileName}
 
-     echo "version=${FFMPEG_VERSION} ${confCmd}" > ${configFileName}
+make &> /dev/null
```
jessp01: Why is the output of make redirected to /dev/null?

build_scripts/build_ffmpeg.sh
```diff
@@ -22,68 +22,43 @@ fi
...
 
-   make &> /dev/null
-popd
+set -e
 exit $RES
```
jessp01: RES is never assigned anything... so it will always exit with 0.

build_scripts/build_addon.sh
```diff
@@ -8,34 +8,56 @@
...
+	echo "unlink ${FFMPEG_SYMLINK}"
+	unlink ${FFMPEG_SYMLINK}
+fi
+
+if [ ! -r ${FFMPEG_SYMLINK} ]; then
+	echo "ln -s ${FFMPEG_PATH} ${FFMPEG_SYMLINK}"
+	ln -s ${FFMPEG_PATH} ${FFMPEG_SYMLINK}
+else
+	echo "failed to unlink ${FFMPEG_PATH}"
+	exit 1
```
jessp01: Always best to use different RCs per case. Since you're already exiting with 1 in the event you didn't get the required args, exit with 2 here.

lilachmaliniak: done

PR #472, opened 2017-06-28
merge to 1.14, QA upgrade scripts update

lilachmaliniak:
merge updated upgradelive and get_bins scripts

### DIFF
--- a/deployment/upgradeLive.sh
+++ b/deployment/upgradeLive.sh
@@ -1,35 +1,37 @@
 #!/bin/bash
 # For upgrade just type ./upgradeLive <version>
 # This uploads NVM
+top -e
 [ -s "$NVM_DIR/nvm.sh" ] && . "$NVM_DIR/nvm.sh" || (echo "nvm not found in $NVM_DIR, this is a must, Exiting!" ; exit 1 )
 cd /opt/kaltura/liveController
-echo updating liveController v.$1
+echo updating liveController v$1
 if [ ! -d "$1" ] ; then
-    if [ ! -L "$latest" ] ; then
-        echo Try to download  https://github.com/kaltura/liveDVR/archive/v$1.tar.gz
-        wget https://github.com/kaltura/liveDVR/archive/v$1.tar.gz
-        echo try to unzip v$1.tar.gz
-        tar -xvzf v$1.tar.gz
-        mv liveDVR-$1 $1
-        cd $1
-        nvm install
-        cp /opt/kaltura/liveController/latest/common/config/configMapping.json /opt/kaltura/liveController/$1/common/config/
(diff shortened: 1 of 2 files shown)


deployment/upgradeLive.sh
```diff
@@ -21,6 +20,7 @@ if [ ! -d "$1" ] ; then
         unlink /opt/kaltura/liveController/latest
         ln -s /opt/kaltura/liveController/$1 /opt/kaltura/liveController/latest
```
gadyaari: I think creation of symbolic link should be at the end of the script after everything has passed properly and before restarting process. 
That way, if something fails it will not overwrite the last version that was linked and was working.

deployment/upgradeLive.sh
```diff
@@ -11,7 +11,6 @@ if [ ! -d "$1" ] ; then
         echo try to unzip v$1.tar.gz
         tar -xvzf v$1.tar.gz
         mv liveDVR-$1 $1
-        bash get_bins "$1"
         cd $1
         nvm install
         cp /opt/kaltura/liveController/latest/common/config/configMapping.json /opt/kaltura/liveController/$1/common/config/
```
gadyaari: We should maintain a configMapping file in the general liveController folder that will enable any version to take the correct configuration without relying on a previous one to exists.
so should be:
cp /opt/kaltura/liveController/configMapping.json /opt/kaltura/liveController/$1/common/config/

deployment/upgradeLive.sh
```diff
@@ -21,6 +20,7 @@ if [ ! -d "$1" ] ; then
         unlink /opt/kaltura/liveController/latest
```
gadyaari: maybe, as you suggested, unlink "latest" as a first step when running the script

deployment/upgradeLive.sh
```diff
@@ -21,6 +20,7 @@ if [ ! -d "$1" ] ; then
...
         ln -s /opt/kaltura/liveController/$1 /opt/kaltura/liveController/latest
         ln -s /opt/kaltura/liveController/latest/serviceWrappers/linux/kLiveController /etc/init.d/kLiveController
+        bash /opt/kaltura/liveController/$1/deployment/get_bins "$1"
```
gadyaari: You have to make sure all the new scripts you create and push have exec premission -> chmod +x
Guy and I saw that the files downloaded from github do not have this permission which will prevent them from running in the script.

deployment/upgradeLive.sh
```diff
@@ -21,6 +20,7 @@ if [ ! -d "$1" ] ; then
...
         /etc/init.d/kLiveController restart
         /etc/init.d/recording_uploder.sh restart
         # also better to call dvr saas config
```
gadyaari: Ron inserted this (the "eco" print line after) not sure why. Might be irrelevant

PR #484, opened 2017-07-11
add installation instructions to readme

noam-arad:


DBezemer:
@noam-arad the 3 build scripts now all use the format <install path> <ffmpeg path> instead of alternating

README.md
```diff
@@ -1,2 +1,30 @@
...
+
+# Installation #
+
+## Pre-requisites ##
+
+* Make sure node is installed
+* Install npm
+* Install nvm (curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.31.2/install.sh | bash)
+* Clone repository
+* C++ compiler
```
jessp01: Note that GCC >= 4.7 is required. This is due to the use of --stg=c++11 in node_addons/FormatConverter/binding.gyp.

README.md
```diff
@@ -1,2 +1,30 @@
 # liveDVR
 DVR implementation built on top of live streams
+
+# Installation #
+
+## Pre-requisites ##
+
+* Make sure node is installed
+* Install npm
+* Install nvm (curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.31.2/install.sh | bash)
```
jessp01: I am confused as to the necessity of NVM. Why not use the official NodeJS RPM/deb packages?

README.md
```diff
@@ -1,2 +1,30 @@
...
+* Wowza + media-server installed and you are able to stream to it.
+
+## Steps to install ##
+
+* Go to repository and run `nvm install;nvm use`
+* Run command `npm install`
+* cd into "build_scripts"
+* Run `./build_ffmpeg.sh <location you want ffmpeg to be> <ffmpeg version>`
+    - NOTE: While this script supports any ffmpeg version we reccomend using "3.0". (no need for 'n' in beginning of version name)
+* Run `./build_ts2mp4_convertor.sh <ffmpeg location> <repository path>/liveRecorder/`
```
jessp01: @lilachmaliniak - Noam's usage here is far more coherent then what the scripts currently have, change the actual usage message accordingly.
Also, whether than "repository path", better use "livedvr clone/source path", which is clearer IMHO.

README.md
```diff
@@ -1,2 +1,30 @@
 # liveDVR
 DVR implementation built on top of live streams
+
+# Installation #
+
+## Pre-requisites ##
+
```
jessp01: Please also add a dep on Python >=2.7

README.md
kalturaguy: in production, QA, it's easier to deploy this way, so we don' need to upgrade node.js anymore, it's automatic using the .nvmrc . file

README.md
```diff
@@ -1,2 +1,30 @@
...
+* Make sure node is installed
```
yossipapi: Specify which node version is required

jessp01: Indeed. Also, it should be properly declared in package.json, for instance:
```
"engines" : { "node" : ">=6.10.2" },
```
See an example in https://github.com/kaltura/pub-sub-server/blob/Lynx-12.20.0/package.json#L5

In addition, see my comment on NVM, why is it needed?

PR #486, opened 2017-07-11
drop shelljs and use os.hostname().

jessp01:


common/utils/hostname.js
```diff
@@ -40,7 +39,7 @@ function getLocalMachineFullHostname() {
     else
     {
         // On Linux
-        var hostnameCmd = shell.exec('hostname -f');
+	var hostnameCmd = os.hostname(); 
         if (hostnameCmd.code !== 0)
```
kalturaguy: currently, we keep the full name in the DB and all API are using the full name, os.hostname  doesn't...

jessp01: I see... if you need the FQDN, then I recommend you use:
https://www.npmjs.com/package/node-fqdn

jessp01:
Hi @kalturaguy,

Seems OK to me. Certainly better than requiring shelljs. Go ahead and merge.

PR #489, opened 2017-07-12
Should not explicitly call python2.7

jessp01:
For one thing, just because the minimum version we require is python
2.7, does not mean we won't be able to work with say Python 2.8 or even
Python 3.

The packaging handles dep declarations and ensures the required python
package is deployed and is first in PATH so that invoking ```python```
shall exec the correct version.

When using RHEL/CentOS and the python27 package (which indeed, the
liveRecorder/serviceWrappers/linux/liveRecorder also relies on, see line
26: . /opt/rh/python27/enable):
/opt/rh/python27/root/usr/bin/python
lrwxrwxrwx. 1 root root 7 Jul 11 10:47 /opt/rh/python27/root/usr/bin/python -> python2
lrwxrwxrwx. 1 root root 9 Jul 11 10:47 /opt/rh/python27/root/usr/bin/python2 -> python2.7

so invoking ```python``` shall result in
/opt/rh/python27/root/usr/bin/python2.7 being called.

On Debian/Ubuntu:
Python 2.7.6
python-minimal: /usr/bin/python
/usr/bin/python2.7

So again, same story.

### DIFF
--- a/liveRecorder/ZombieEntryCatchers.py
+++ b/liveRecorder/ZombieEntryCatchers.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python2.7
+#!/usr/bin/env python
 import time
 import logging.handlers
 import os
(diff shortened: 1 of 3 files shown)


kalturaguy:
already merged with #491

PR #498, opened 2017-07-18
PLAT-7663: directories of if incoming should be sorted by creation time

noam-arad:


liveRecorder/Tasks/TaskRunner.py
```diff
@@ -97,8 +97,8 @@ def schedule_job(self):
             time.sleep(1)
 
     def move_and_add_to_queue(self, src_dir):
-
-        for directory_name in os.listdir(src_dir):
+        file_list = self.getSorterFileList(src_dir)
```
kalturaguy: would add a comment why we sort (avoid starvation...)

liveRecorder/Tasks/TaskRunner.py
```diff
@@ -227,3 +227,17 @@ def safe_move(self, src, dst):
 
         except Exception as e:
             self.logger.error("Failed to move %s to %s : %s \n %s", src, dst, str(e), traceback.format_exc())
+    def getSorterFileList(self, src_dir):
+        file_list = os.listdir(src_dir)
+        file_list_with_ctime = []
+        for path in file_list:
```
kalturaguy: usually i prefer:
file_list_with_ctime = [ (path,os.stat(src_dir + '/' + path).st_ctime) for path in file_list]

liveRecorder/Tasks/TaskRunner.py
```diff
@@ -227,3 +227,17 @@ def safe_move(self, src, dst):
...
+            full_path = src_dir + '/' + path
```
kalturaguy: use os.path.join for joining paths (safer in case of extra/missing /)

PR #502, opened 2017-07-21
Fixes

jessp01:


build_scripts/build_addon.sh
```diff
@@ -80,12 +80,12 @@ pushd ${ADDON_PATH}
 	echo "Start node-gyp build. ${GYP_DEBUG}"
 	node-gyp build ${GYP_DEBUG} -v
 
-    if [ -r "build/${BUILD_CONF}/lib.target/${FORMAT_CONVERTER_BIN}"  ]; then
-		echo "cp build/${BUILD_CONF}/lib.target/${FORMAT_CONVERTER_BIN} ${PRODUCT_ROOT_PATH}/bin/FormatConverter.node${DEBUG_EXT}"
```
kalturaguy: cannot merge this, since this was requested by David

liveRecorder/serviceWrappers/linux/liveRecorder
```diff
@@ -29,8 +29,8 @@ SCRIPT_DIR=$(dirname $(readlink -f $0))
 APPDIR=${SCRIPT_DIR}/../../
 NAME=${APP_NAME}
 APPLICATION_PATH=${APPDIR}/main.py
-LOG_DIR="/var/log/${APP_NAME}"
-PID_DIR="/var/run"
+LOG_DIR="@LOG_DIR@/${APP_NAME}"
```
kalturaguy: this will break production/QA

serviceWrappers/linux/kLiveController
```diff
@@ -57,7 +57,7 @@ loadNvm() {
 
 start() {
     [ -d ${LOG_DIR} ] || mkdir -p ${LOG_DIR}
-    loadNvm
+    #loadNvm
```
kalturaguy: this will break QA/Production where we install node via NVM

build_scripts/build_addon.sh
jessp01: @DBezemer, please comment.

liveRecorder/serviceWrappers/linux/liveRecorder
jessp01: QA should and usually do use the same packages I release.
As for SaaS - Guy E. should add the replacement of tokens in his deployment script or else, he can push his own version of this script to the saas-config repo, that's what it is there for.
@GElkayam - your thoughts please.

serviceWrappers/linux/kLiveController
jessp01: Question is why. Like I wrote, NodeJS provides perfectly usable RPM and deb packages in their official repos. See:
https://nodejs.org/en/download/package-manager/

kalturaguy: Currently prodit does work with rpm. And i cannot force them to that. Since they are pretty bust team and the ROI of that is negative
You can put a env. Val  that can define if to use deault log folder and if to use nvm so you are comptaible with other enviroments and deployment methods

jessp01: Installing the official RPM/deb packages is better for Prod. too and it would take exactly 1 minute to change it but like I said before, their considerations are theirs and should not affect the OP packages. 

If @GElkayam wishes to continue to use NVM, he can. Here are several options:
- Use a different init script. Usually SaaS specific scripts go into either the 'saas-' repos or into ITProd's own repo. 

- We can have two sets of init scripts in the liveDVR repo, though personally, I think that's not a good approach.

- Alternatively, and perhaps, best, you can change the script so that:
```
KLIVE_CONTROLLER_PREFIX="/opt/kaltura/liveController/latest"
LOG_DIR="/var/log/liveController"
PID_DIR="/var/run"
```
is removed and more specific ENV vars like LIVEDVR_LOG_DIR and LIVEDVR_RUN_DIR are used and then I can set these and KLIVE_CONTROLLER_PREFIX in say /etc/default/livedvr [Debian/Ubuntu] and /etc/sysconfig/livedvr [RHEL/CentOS] and IT Prod. can do the same on their system.
and thr loadNVM function can still check for:
```
 if [ -s "${NVM_DIR}/nvm.sh" ]; then
```
and source it if it exists but not exit if it does not. Once again, requiring it is wrong since your code can work with any valid NodeJS of version 6 and higher, it does not depend on anything that is provided by NVM exclusively.

zoharbabin: @kalturaguy @yossipapi @GElkayam - the official way to install nodejs is by using their official rpm and deb packages.
Debina/Ubuntu: https://nodejs.org/en/download/package-manager/#debian-and-ubuntu-based-linux-distributions
RedHat/Fedora: https://nodejs.org/en/download/package-manager/#enterprise-linux-and-fedora
As a design guideline, when creating rpm/deb packages it is best practice to use other rpm/deb packages whenever available as dependancies. That allows us to continue using the same system for tracking changes, dependancies, and overall provide a more reliable way to install the platform.
NVM requires further building of non rpm/deb packages, it requires python, and overall introduces further dependancies that will complicate the platform deployment and make the system prone for other installation errors that can't leverage the standard rpm/deb install mechanism.

For SaaS we manage the environment, so there's very little risk. But when deployed and managed by others on other environment, we need the install packages to be as fail safe as possible, and thus minimizing the list of dependancies is critical.

Thanks!

liveRecorder/serviceWrappers/linux/liveRecorder
GElkayam: I'm fine with saving this file as .template and the original file in the saas-config, so the rpm installation will create from template.
I'm also fine with loading the default values for environment variables from /etc/default, if a file exists.

serviceWrappers/linux/kLiveController
GElkayam: To solve the issue, and make everyone happy, we can keep a variable for USE_NVM or LOAD_NODE_VIA_NVM (true/false).
In production we choose to use nvm so we know that the version of node is exactly as the version that was approved by RnD, and planned to be used.
When the product started (LiveDVR) it used node 0.10.40, and ran in the same time with then new LiveController which needed Node 6.3.0, so using nvm was the way to go.
if in future versions we will not need to run different node versions on the same server, and do not fear upgrades that will change code behaviuor, we can go with official installed rpm (deb) packages.
IMHO, nvm is the way to go, as it defines in code the specific node versions, but I'm no expert on OnPrems installation standards.

deployment/upgradeLive
```diff
@@ -120,10 +120,13 @@ function downgrade() {
 }
 
 function stop() {
+   echo "stopping liveController..."
```
jessp01: We should check for -x /etc/init.d/liveRecorder not -r, also, since we're checking, we should check for both.
Why do we echo "stopping ..." here? the "stopping" output should be done in the control scripts themselves, not here.

deployment/upgradeLive
```diff
@@ -133,13 +136,13 @@ function start() {
     fi
     # start nginx (web server), liveController and liveRecorder
     echo "starting liveController..."
-    /etc/init.d/kLiveController start
+    /etc/init.d/kLiveController restart
```
jessp01: Same as above.

liveRecorder/serviceWrappers/linux/liveRecorder
```diff
@@ -24,13 +24,24 @@
...
 APPDIR=${SCRIPT_DIR}/../../
 NAME=${APP_NAME}
 APPLICATION_PATH=${APPDIR}/main.py
-LOG_DIR="/var/log/${APP_NAME}"
-PID_DIR="/var/run"
+LOG_DIR=@LOG_DIR@
+PID_DIR=@RUN_DIR@
+
+#default values
+if [ "${LOG_DIR}" = "@LOG_DIR@" ]; then
```
jessp01: This doesn't make any sense. Either your transform this file into a template file from which the init will be generated or you use ENV vars as I explained in my comment here:
https://github.com/kaltura/liveDVR/pull/502/files/41168d211ee18133c668f72f7f9a00efb892d51e#r128896835

serviceWrappers/linux/kLiveController
```diff
@@ -23,36 +23,55 @@
 ###
 source /etc/rc.d/init.d/functions
 
+LOG_DIR=@LOG_DIR@
+PID_DIR=@RUN_DIR@
+USE_NVM=@USE_NVM@
+KLIVE_CONTROLLER_PREFIX="@KLIVE_CONTROLLER_PREFIX@"
+
+#default values
+if [ "${LOG_DIR}" = "@LOG_DIR@" ]; then
```
jessp01: See https://github.com/kaltura/liveDVR/pull/502/files#r129044990

serviceWrappers/linux/kLiveController
```diff
@@ -23,36 +23,55 @@
...
+
+
 NAME="kLiveController"
-KLIVE_CONTROLLER_PREFIX="/opt/kaltura/liveController/latest"
-NODE_PATH="$KLIVE_CONTROLLER_PREFIX/node_modules"
-FOREVER="$NODE_PATH/forever/bin/forever"
-APPLICATION_PATH="$KLIVE_CONTROLLER_PREFIX/lib/App.js"
-LOG_DIR="/var/log/liveController"
-PID_DIR="/var/run"
+NODE_PATH="${KLIVE_CONTROLLER_PREFIX}/node_modules"
```
jessp01: This is wrong because on OP ENVs, connection to the outside world is often not allowed and thus, NPMs cannot be retrieved from http://npmjs.com. We therefore package each such NPM module as an RPM and install it onto /usr/lib/npm_modules, and so, NODE_PATH should also be a token.

serviceWrappers/linux/kLiveController
```diff
@@ -23,36 +23,55 @@
...
 loadNvm() {
-    # This loads nvm
-    # must cd/pushd due to a bug in /root/.nvm/nvm.sh call to nvm_find_nvmrc()  
-    pushd ${KLIVE_CONTROLLER_PREFIX}
-    [ -z "${NVM_DIR}" ] && NVM_DIR="${HOME}/.nvm"
-    if [ -s "${NVM_DIR}/nvm.sh" ]; then
-        source "${NVM_DIR}/nvm.sh"
-        if [ -s ".nvmrc" ] ; then
-            nvm use || nvm install
+    if [ "${USE_NVM}" = "1" ]; then
```
jessp01: Why not check if USE_NVM is set outside of the function and not call it if it isn't?

kalturaguy: better to check in one place than 10....

etameran:
Jess, after diving into the two different issues raised by the recent pull request, I've suggested the following:

1. We will use NVM as this is the flexible tool we selected for our SaaS env. , You can add a condition around its inclusion if you want to avoid its use in onPrem/CE, 
2. As for template params. In production we don't reinstall the package but rather resync the code. In order to provide both options, I've instructed the team to create an "default_vars.sh" file which will set default vars and if they don't exist will use the template vars. Moving forward, we will ensure variables ($var_name) will be used instead of constants paths however we will keep defaults as they are better for our use in the SaaS env.

jessp01:
Hi Eran,

In regards to NVM, I'm fine with that, I suggested it myself in previous comments and discussions. 
See my last note on that here:
https://github.com/kaltura/liveDVR/pull/502#pullrequestreview-51781441

As for ENV vars pointing to different paths [log, run, etc], please see my comment here:
https://github.com/kaltura/liveDVR/pull/502/files/41168d211ee18133c668f72f7f9a00efb892d51e#r128896835
What I don't think is right is the current state of affairs, to wit:
```
if [ "${LOG_DIR}" = "@LOG_DIR@" ]; then
```

Thanks,

etameran:
I understand but since we don't have saas tag versions for shell scripts I rather go with the default vars

Eran

Sent from my iPhone

On 24 Jul 2017, at 23:08, Jess Portnoy <notifications@github.com<mailto:notifications@github.com>> wrote:


Hi Eran,

In regards to NVM, I'm fine with that, I suggested it myself in previous comments and discussions.
See my last note on that here:
#502 (review)<https://github.com/kaltura/liveDVR/pull/502#pullrequestreview-51781441>

As for ENV vars pointing to different paths [log, run, etc], please see my comment here:
https://github.com/kaltura/liveDVR/pull/502/files/41168d211ee18133c668f72f7f9a00efb892d51e#r128896835
What I don't think is right is the current state of affairs, to wit:

if [ "${LOG_DIR}" = "@LOG_DIR@" ]; then


Thanks,

PR #504, opened 2017-07-27
merge to 1.16 fix VOD upload for recording zombie job

lilachmaliniak:
fix upload VOD of recording job created by zombie recover cron


lilachmaliniak:
small fix that enables processing of recovered zombie recordings.

PR #508, opened 2017-07-30
Merge Fixes to 1.17

lilachmaliniak:
Jess’s updates
+ new template files for liveRecorder and kLiveController service
wrappers.


serviceWrappers/linux/kLiveController
```diff
@@ -21,48 +21,71 @@
...
-        exit 1
+        popd
     fi
-    popd
 }
 
 start() {
-    [ -d ${LOG_DIR} ] || mkdir -p ${LOG_DIR}
+    echo "starting liveController..."
+    [ -d ${LOG_DIR}/liveController ] || mkdir -p ${LOG_DIR}/liveController
```
kalturaguy: liveController is concatenated to  LOG_DIR all around... so maybe it's better that the LOG_DIR will point to the this path

lilachmaliniak: done

liveRecorder/serviceWrappers/linux/liveRecorder
```diff
@@ -22,18 +22,42 @@
...
+fi
+
+# these lines added especially for QA
+if [ ! -r ${KLIVE_RECORDER_PREFIX} ]; then
+    SCRIPT_DIR=$(dirname $(readlink -f $0))
+    KLIVE_RECORDER_PREFIX=${SCRIPT_DIR}/../../
+    echo "KLIVE_RECORDER_PREFIX=${KLIVE_RECORDER_PREFIX}"
+fi
+
+LOG_DIR="${KLIVE_LOG_DIR}/${NAME}"
```
kalturaguy: where is KLIVE_LOG_DIR defined?

serviceWrappers/linux/kLiveController
kalturaguy: still same....

serviceWrappers/linux/kLiveController
```diff
@@ -21,48 +21,71 @@
 ###
 # extend callable functions from "functions" script
 ###
+NAME=kLiveController
+LOG_DIR=/var/log
```
jessp01: LOG_DIR should not be a literal, plus, it is reassigned in line 56. PID_DIR should also not be a literal, it is not reassigned later, though.

serviceWrappers/linux/kLiveController
```diff
@@ -21,48 +21,71 @@
...
 loadNvm() {
-    # This loads nvm
-    # must cd/pushd due to a bug in /root/.nvm/nvm.sh call to nvm_find_nvmrc()  
-    pushd ${KLIVE_CONTROLLER_PREFIX}
-    [ -z "${NVM_DIR}" ] && NVM_DIR="${HOME}/.nvm"
-    if [ -s "${NVM_DIR}/nvm.sh" ]; then
-        source "${NVM_DIR}/nvm.sh"
-        if [ -s ".nvmrc" ] ; then
-            nvm use || nvm install
+    if [ "${USE_NVM}" = "1" ]; then
```
jessp01: This condition should be OUTSIDE the function, is USE_NVM is not set to 1, loadNvm() should not be called.

liveRecorder/serviceWrappers/linux/liveRecorder
lilachmaliniak: updated

serviceWrappers/linux/kLiveController
lilachmaliniak: Jess,
Eran & Guy E. agreed that there will be de literal values for LOG_DIR etc. and that we won't add defaults to live saas repository. That means that there are values for that those parameters defined in the template but if default file exists they are replaced. In production there won't be defaults file so they won't be replaced.

lilachmaliniak: Eran=Eran E.

serviceWrappers/linux/kLiveController
lilachmaliniak: done

serviceWrappers/linux/kLiveController
jessp01: Sorry but I don't understand. What's so hard about:
```
if [ -z "$LIVE_LOG_DIR" ];then
   LIVE_LOG_DIR=/var/log
fi
```
This way, if LIVE_LOG_DIR was defined OUTSIDE the script, it shall be used, otherwise /var/log will be used.
This solution will work for both OP and SaaS and will require no replacements inside init or other files.

PR #509, opened 2017-08-01
merge to 1.17 PLAT-7109 recording lost when flavors has changed between sessions

lilachmaliniak:


lib/recording/RecordingManager.js
```diff
@@ -331,6 +350,11 @@ class RecordingManager {
                         return memo + flavorChunks
                     }, 0);
                     //todo should also print warning if # of chunks is not equal between all flavors!
+                    if (isNaN(recordingEntrySession.startDuration)) {
+                        logger.warn(`[${entryId}-${recordingEntrySession.recordedEntryId}][stopRecording] Recording didn't start, do nothing.`);
+                        delete this.recordingList[entryId];
+                        return qio.write(doneFilePath, 'done')
```
gadyaari: return Q.resolve()

lib/BackendClient.js
```diff
@@ -500,7 +500,7 @@ var BackendClient = (function(){
     };
 
     BackendClient.resetRecordingEntry = function(entryObj) {
-        logger.debug(`[${entryObj.entryId}] Recording reached max duration, reset recordedEntryId`);
+        logger.debug(`[${entryObj.entryId}] Recording reached max duration or provision difference from previous r, reset recordedEntryId`);
```
lilachmaliniak: fixed

lib/recording/RecordingManager.js
```diff
@@ -186,8 +202,11 @@ class RecordingManager {
         let recordingSession = this.recordingList[entryId];
         if (!recordingSession.mp4FilesQueueLock) {
             recordingSession.mp4FilesQueueLock = true;
-            logger.info(`[${entryId}] Recording entry ${this.recordingList[entryId].recordedEntryId} has passed the limit of recording duration - ${this.recordingMaxDurationInMs/1000} seconds. Split session`);
-            return recordingSession.handleReachedMaxRecordingDuration()
+            logger.info(`[${entryId}] Recording entry ${this.recordingList[entryId].recordedEntryId} has passed the limit of recording duration - ${this.recordingMaxDurationInMs / 1000} seconds. Split session`);
+            return this.splitRecordedEntry(recordingSession.entryObject)
```
lilachmaliniak: fixed

lib/recording/RecordingManager.js
lilachmaliniak: done

lib/recording/RecordingManager.js
```diff
@@ -375,7 +399,65 @@ class RecordingManager {
         }
     }
 
+    verifyStreamMatchPreviousFlavorSet(entryObj) {
```
lilachmaliniak: moved

lib/recording/RecordingEntrySession.js
```diff
@@ -132,20 +125,21 @@ class RecordingEntrySession {
         return Q.resolve();
     }
 
-    handleReachedMaxRecordingDuration() {
-        if (this.entryObject.serverType == kalturaTypes.KalturaEntryServerNodeType.LIVE_PRIMARY) {
+    splitRecordedEntry(entryObject) {
```
lilachmaliniak: done and will be verified running 24-7 flow

lib/recording/RecordingManager.js
```diff
@@ -97,7 +97,7 @@ class RecordingManager {
                                 return recordingEntrySession.splitRecordedEntry(entryObj)
                                     .then(() => {
                                         logger.info(`[${entryObj.entryId}-${entryObj.recordedEntryId}] after split, [${recordingEntrySession.recordedEntryId}] replaced by [${entryObj.recordedEntryId}]`);
-                                        return this.getRecordingEntrySession(entryObj);
+                                        return this.getRecordingEntrySession(entryObj, recordingEntrySession.mp4FilesQueueLock);
```
gadyaari: mp4FilesQueueLock is just a boolean lock
should be mp4FilesQueue

lib/SessionManager.js
```diff
@@ -10,22 +10,24 @@ var _ = require('underscore');
...
+        logger.debug("[%s] Entry last updated at %s; time passed = %d sec; Grace period duration: %d seconds. Time to stop session.", entryId, lastModificationTime, (currentTime - lastModificationTime)/1000, flavorDownloaderTeardownInterval/1000);
+        return ((currentTime - lastModificationTime) >= flavorDownloaderTeardownInterval);
+    }
+
+    function isTimeToCreateNewSession(lastModificationTime, entryId) {
+        logger.debug("[%s] Checking if this is a new session", entryId);
         var currentTime = new Date();
-        logger.debug("[%s] Entry was last live at %s; time passed = %d sec; Grace period duration: %d seconds", entryId, lastModificationTime, (currentTime - lastModificationTime)/1000, timeToWait/1000);
-        return ((currentTime - lastModificationTime) >= timeToWait);
+        logger.debug("[%s] Entry was last live at %s; time passed = %d sec; Grace period duration: %d seconds. Need to create new live session.", entryId, lastModificationTime, (currentTime - lastModificationTime)/1000, sessionDuration/1000);
```
gadyaari: log line is not inline with the function's name and functionality. 
Sentence states that we Need to create a new live session but what we are actually doing is Checking if we need to create one - which means the answer can be NO but yet we still stated that we need to...

lib/SessionManager.js
```diff
@@ -10,22 +10,24 @@ var _ = require('underscore');
...
-        }
-        else {
-            logger.debug("[%s] Checking if to call 'stop' for entry", entryId);
-            timeToWait = config.get('flavorDownloaderTeardownInterval');
-        }
+    function isTimeToStopSession(lastModificationTime, entryId) {
+        logger.debug("[%s] Checking if this is a new session", entryId);
+        var currentTime = new Date();
+        logger.debug("[%s] Entry last updated at %s; time passed = %d sec; Grace period duration: %d seconds. Time to stop session.", entryId, lastModificationTime, (currentTime - lastModificationTime)/1000, flavorDownloaderTeardownInterval/1000);
+        return ((currentTime - lastModificationTime) >= flavorDownloaderTeardownInterval);
```
gadyaari: Same as comment below

lib/utils/fs-utils.js
```diff
@@ -59,8 +59,53 @@ function existsAndNonZero(path) {
         });
 }
 
+function isSameFlavorSet(flavorsPath, flavorsSet, excludeDirs, pattern) {
+    return qio.listTree(flavorsPath,
```
kalturaguy: same comment as before... you have the playlist generator.... why  you need to look at the disk?

lib/utils/fs-utils.js
```diff
@@ -59,8 +59,53 @@ function existsAndNonZero(path) {
...
+                return false;
+            }
+            if (pattern.test(dirName)) {
+                return true;
+            } else {
+                return false;
+            }
+        })
+        .then((flavorsDirs)=> {
+            let flavorNames = _.map(flavorsDirs, function(flavorDir) {
```
kalturaguy: the naming is confusing... flavorNames should be called currentFlavors . or previousFlavors

lib/recording/RecordingEntrySession.js
```diff
@@ -132,20 +189,21 @@ class RecordingEntrySession {
         return Q.resolve();
     }
 
-    handleReachedMaxRecordingDuration() {
-        if (this.entryObject.serverType == kalturaTypes.KalturaEntryServerNodeType.LIVE_PRIMARY) {
+    splitRecordedEntry(entryObject = this.entryObject) {
```
kalturaguy: this is not good, please don't give such default properties, it's bit unpredictable what is the value

lib/recording/RecordingEntrySession.js
```diff
@@ -369,16 +427,72 @@ class RecordingEntrySession {
 
     }
 
+ /*   verifyStreamMatchPreviousFlavorSet(newEntryObj) {
```
kalturaguy: if this method is not used, then delete it

lib/recording/RecordingManager.js
```diff
@@ -50,133 +50,83 @@ class RecordingManager {
...
-                    return Q.reject(err);
+            return this.getRecordingEntrySession(entryObj, mp4FilesQueue)
+                .then((recordingObj)=> {
+                    return recordingObj.session.verifyFlavorSetMatchPlaylistFlavors(entryObj)
+                        .then((flavorsMatch) => {
+                            if (!flavorsMatch) {
+                                return recordingObj.session.splitRecordedEntry(entryObj)
+                                    .then(() => {
+                                        logger.info(`[${entryObj.entryId}-${entryObj.recordedEntryId}] after split, [${recordingObj.session.recordedEntryId}] replaced by [${entryObj.recordedEntryId}]`);
+                                        return this.getRecordingEntrySession(entryObj);
```
kalturaguy: aren't you missing here the mp4FilesQueue?
return this.getRecordingEntrySession(entryObj,mp4FilesQueue);

lib/recording/RecordingEntrySession.js
```diff
@@ -369,16 +427,72 @@ class RecordingEntrySession {
...
+                return _.map(playlistJSON.sequences, (sequence) => {
+                    return sequence.id;
+                });
+            })
+            .then((flavorsFromPlaylist)=> {
+                let newFlavorNames = _.map(newEntryObj.flavorsObjArray, function(newFlavorDir) {
+                    return newFlavorDir.name;
+                });
+                // check if flavors match current provisioned flavors
+                if (flavorsFromPlaylist.length != newFlavorNames.length) {
```
kalturaguy: duplicated logic, below logic is enough...   _.difference should work if size changes too

lib/recording/RecordingEntrySession.js
```diff
@@ -369,16 +427,72 @@ class RecordingEntrySession {
...
+    verifyFlavorSetMatchPlaylistFlavors(newEntryObj) {
+        let playlistFilePath = path.join(this.recordingSessionPath, 'playlist.json');
+        return qio.read(playlistFilePath)
+            .catch((err) =>{
+                if (err instanceof Error && err.code !== 'ENOENT') {
+                    this.logger.error(`Failed to read [${this.recordingSessionPath}]. Error: ${ErrorUtils.error2string(err)}`);
+                    return Q.reject(err);
+                }
+                else
+                    return Q.resolve();
```
kalturaguy: how will this work if the folder doesn't exists? (new session)
you pass Q.Resolve, and then next then you try to parse it?

lib/utils/fs-utils.js
```diff
@@ -59,8 +59,53 @@ function existsAndNonZero(path) {
         });
 }
 
+function isSameFlavorSet(flavorsPath, flavorsSet, excludeDirs, pattern) {
```
kalturaguy: you don't use this function anymore  please delete it

lib/recording/RecordingManager.js
```diff
@@ -50,133 +50,83 @@ class RecordingManager {
...
+                                    })
+                            } else {
+                                return recordingObj;
+                            }
+                        })
+                        .then((recordingObj) => {
+                            if (!recordingObj.started) {
```
kalturaguy: not sure what is the purpose of started here? why it worked before and not it's required

lib/recording/RecordingEntrySession.js
lilachmaliniak: not exactly... the difference returns the set of items from array (other) that are not on the right so _.difference( [32], [32,33]) will return an empty array, thus wrong to decide that flavors didn't change. Question is whether this is possible without changing transcoding profile: [32,33] => [32,34] if this is impossible then it is enough to compare arrays size. 
_.difference(array, *others) 
returns the values from array that are not present in the other arrays.

lib/recording/RecordingManager.js
lilachmaliniak: In the past Ron returned null instead of recordedEntrySession and that was the indicator that a recording session exists. I replaced it with the isStarted flag

lib/recording/RecordingEntrySession.js
kalturaguy: then use _. intersection 

 _. intersection(flavorsFromPlaylist, newFlavorNames).length!=flavorsFromPlaylist.length

lib/recording/RecordingManager.js
kalturaguy: then rename the isStarted to newSession/appendSession?

lib/recording/RecordingManager.js
```diff
@@ -331,6 +284,11 @@ class RecordingManager {
                         return memo + flavorChunks
                     }, 0);
                     //todo should also print warning if # of chunks is not equal between all flavors!
+                    if (isNaN(recordingEntrySession.startDuration)) {
```
kalturaguy: please add remark here what is the use case

lib/recording/RecordingManager.js
```diff
@@ -50,133 +50,83 @@ class RecordingManager {
...
+        }
+        if (!this.recordingList[entryId]) {
+            logger.info(`[${entryObj.entryId}-${entryObj.recordedEntryId}] Create new recording session`);
+            let recordingEntrySession = new RecordingEntrySession(entryObj);
+            this.recordingList[entryId] = recordingEntrySession;
+            return Q.resolve({session:recordingEntrySession, started:false});
+        }
+        else if (this.recordingList[entryId].recordedEntryId != entryObj.recordedEntryId) {
+            logger.info(`[${entryObj.entryId}-${entryObj.recordedEntryId}] New recorded entryId has been generated. Delete existing session and start new one`);
+            return this.stopRecording(entryObj.entryId)
```
kalturaguy: it doesn't sound logic to me that the function that is called getRecordingEntrySession is doing stopRecording

lib/recording/RecordingEntrySession.js
```diff
@@ -369,16 +426,51 @@ class RecordingEntrySession {
 
     }
 
+    verifyFlavorSetMatchPlaylistFlavors(newEntryObj) {
+        let playlistFilePath = path.join(this.recordingSessionPath, 'playlist.json');
+        return qio.read(playlistFilePath)
+            .then((data) => {
+                let playlistJSON = JSON.parse(data);
```
kalturaguy: can't we use the playListgenerator class to laod the json?
if we decide to change the format in the future, this code will break

lib/recording/RecordingManager.js
```diff
@@ -50,133 +50,84 @@ class RecordingManager {
...
-                logger.debug(`[${entryObj.entryId}-${entryObj.recordedEntryId}] startNewSession`);
-                return qio.makeTree(recordingEntrySession.recordingSessionPath)
-                    .then(()=>{
-                        return makeFlavorDirectories(recordingEntrySession)
-                    })
-                    .then(()=>{
-                        return recordingEntrySession.accessibleRecording()
-                    })
-            };
+    getRecordingEntrySession(entryObj, mp4FilesQueue = null) {
```
kalturaguy: please rename to createRecordingEntrySession

lib/recording/RecordingManager.js
```diff
@@ -50,133 +50,84 @@ class RecordingManager {
...
+                                    })
+                                    .then((mp4FilesQueue)=> {
+                                        return this.getRecordingEntrySession(entryObj, mp4FilesQueue)
+                                    })
+                            } else {
+                                return sessionObj;
+                            }
+                        })
+                        .then((sessionObj) => {
+                            if (sessionObj.newSession) {
```
kalturaguy: at the past you called startRecording even for existing session...

lib/SessionManager.js
```diff
@@ -10,22 +10,33 @@ var _ = require('underscore');
...
-        var timeToWait;
-        if (isFromDisk) {
-            logger.debug("[%s] Checking if this is a new session", entryId);
-            timeToWait = config.get('sessionDuration');
-        }
-        else {
-            logger.debug("[%s] Checking if to call 'stop' for entry", entryId);
-            timeToWait = config.get('flavorDownloaderTeardownInterval');
+    function isTimeToStopSession(lastModificationTime, entryId) {
+        logger.debug("[%s] Checking if this is a new session", entryId);
```
lilachmaliniak: updated phrasing

lib/SessionManager.js
```diff
@@ -10,22 +10,33 @@ var _ = require('underscore');
...
+    }
+
+    function isTimeToCreateNewSession(lastModificationTime, entryId) {
+        logger.debug("[%s] Checking if this is a new session", entryId);
         var currentTime = new Date();
-        logger.debug("[%s] Entry was last live at %s; time passed = %d sec; Grace period duration: %d seconds", entryId, lastModificationTime, (currentTime - lastModificationTime)/1000, timeToWait/1000);
-        return ((currentTime - lastModificationTime) >= timeToWait);
+        var elapsedTimeMS = currentTime - lastModificationTime;
+        logger.debug(`[${entryId}] Entry was last live at ${lastModificationTime}; time passed = ${(elapsedTimeMS/1000).toFixed(2)} sec; Grace period: ${sessionDuration/1000} seconds`);
+        if (elapsedTimeMS >= sessionDuration) {
```
lilachmaliniak: 1. renamed
2. added method

lib/recording/RecordingEntrySession.js
lilachmaliniak: replaced read playlist with method in playlistGenerator

lib/recording/RecordingManager.js
lilachmaliniak: why? in some of the cases nothing is created since session exists

PR #510, opened 2017-08-05
1.16-plat 6362: update healthness to backend

gadyaari:


lib/entry/LiveEntry.js
```diff
@@ -212,8 +204,15 @@ LiveEntry.prototype.update = function(diagnosticsObj) {
...
-            this.diagnosticsAnalyzer.runFullDiagnosticsAnalysis(diagnosticsObj);
+            _.each(this.flavorsDownloaders, (f)=> {
+                f.setEncodedBitrate(this.adapterDiagnosticsObj.outputs[f.flavor]);
+            });
+            let entryInfo = {
+                'source' : this.adapterDiagnosticsObj,
+                'flavors' : JSON.stringify(this.flavorsDownloaders),
+                'playlistGenerator' : JSON.stringify(this.playlistGenerator)
+            };
+            this.diagnostics.handleDiagnosticsReport(entryInfo);
```
kalturaguy: note that here you are stringfing the flavors object and playlist generator.... this is not good, you should just call toJson, otherwsie you will work with strings and not object?

lib/Diagnostics/Diagnostics.js
```diff
@@ -2,23 +2,62 @@
...
+        this._runDiagnosticsReport.call(this);
     }
 
     toJSON() {
         return { "internalAlerts" : this._internalAlerts, "externalAlerts" : Array.from(this._externalAlerts) };
     }
-    
+
+    _createAlertsAnalyzer(entryId) {
+        let an = new DiagnosticsAlertsAnalyzer(entryId, this.alertWindowSizeInMSec);
```
kalturaguy: you need to put DiagnosticsAlertsAnalyzer instance as a class member, otherwise GC will delete this object...

this.an=new DiagnosticsAlertsAnalyzer(entryId, this.alertWindowSizeInMSec);

common/config/config.json.template
```diff
@@ -57,6 +57,7 @@
 
   },
   "diagnostics": {
+      "diagnosticsReportIntervalRunTimeInSec": 20,
```
kalturaguy: will put here even 1 min.

lib/Diagnostics/Diagnostics.js
```diff
@@ -2,23 +2,62 @@
...
+        // Enlist on events
+        // TODO: Check with Eran/Amir that this is the correct way to enlist on event!!!
+        an.on('pushAlert', this.pushAlert.bind(this));
```
kalturaguy: general comment: don't like the pushalert name.... maybe reportAlert?

lib/Diagnostics/Diagnostics.js
```diff
@@ -29,8 +68,23 @@ class Diagnostics {
         this._logger.debug('pushAlert. Alert: %s', diagnosticsAlertObj.name);
     }
 
-    clearOldAlerts() {
-        let that = this;
+    removeAlert(key) {
+        // If hasKey is for an alert of type 'EntryRestartedAlert' && alert exists
+        // (Only alert of type EntryRestartedAlert will have hashKey '100_')
+        if (key.includes(diagAlertsErrorCodes.EntryRestartedAlert + '_') &&
```
kalturaguy: why do we need this, what's so special with EntryRestartedAlert?

lib/entry/LiveEntry.js
gadyaari: done

lib/Diagnostics/Diagnostics.js
gadyaari: done

common/config/config.json.template
gadyaari: done. Need further discussion

lib/BackendClient.js
```diff
@@ -433,6 +439,35 @@ var BackendClient = (function(){
...
+        logger.debug(`[${entryId}] Calling beacon.add, Tag: [${tag}]`);
+
+        let KalturaBeacon = new kalturaVO.KalturaBeacon();
+        KalturaBeacon.relatedObjectType = kalturaTypes.BeaconObjectTypes.ENTRY_SERVER_NODE_BEACON;
+        KalturaBeacon.eventType = tag;
+        KalturaBeacon.objectId = entryId;
+        KalturaBeacon.privateData = JSON.stringify(data);
+
+        //Impersonate call to the entry's partnerId to execute the call in the partner's context
+        client.setPartnerId(partnerId);
```
kalturaguy: this is bad... since from this point onward ALL API requests will be done with this specific partnerId...

lib/Diagnostics/Diagnostics.js
```diff
@@ -2,35 +2,119 @@
...
+                this._logger.debug('Diagnostics Module stopped');
+            })
+    }
+
+    insertAlert(diagnosticsAlertObj) {
+        this._logger.debug(`[insertAlert] Alert: ${diagnosticsAlertObj.name}`);
         if (diagnosticsAlertObj.hashKey) {
-            this._externalAlerts.set(diagnosticsAlertObj.hashKey, diagnosticsAlertObj);
+            this.reportedAlerts.set(diagnosticsAlertObj.hashKey, diagnosticsAlertObj);
+            return this._updateStreamHealth.call(this)
```
kalturaguy: won't it be better to mark a flag as dirty and report only once at the end of all alerts calculations?

lib/Diagnostics/Diagnostics.js
```diff
@@ -44,12 +128,134 @@ class Diagnostics {
...
+        // (2) If one of the inputs was removed or another has been added -> send beacon
+        // (3) If an input was changed and now has a different sessionId -> send beacon
+        let tempStaticData = JSON.parse(JSON.stringify({ 'inputs': metaData.source.inputs, 'playlistGenerator': metaData.playlistGenerator }));
+        _.each(tempStaticData.inputs, i => {
+           delete i.IOPerformance;
+           delete i.Encoder.syncPTSData;
+           delete i.Properties.timeRunningSeconds;
+        });
+
+        if (!this.staticMetaData.data) {
```
kalturaguy: you can do it with one line: this.staticMetaData.modified = !_.isEqual(tempStaticData, this.staticMetaData.data)

lib/BackendClient.js
```diff
@@ -433,6 +439,37 @@ var BackendClient = (function(){
...
+    function sendBeacon(entryId, clientPartnerId, data, beaconTag) {
+        let tag = beaconPrefix + '_' + beaconTag;
+        let deferred = Q.defer();
+        logger.debug(`[${entryId}] Calling beacon.add, Tag: [${tag}]`);
+
+        let KalturaBeacon = new kalturaVO.KalturaBeacon();
+        KalturaBeacon.relatedObjectType = kalturaTypes.BeaconObjectTypes.ENTRY_SERVER_NODE_BEACON;
+        KalturaBeacon.eventType = tag;
+        KalturaBeacon.objectId = entryId;
+        KalturaBeacon.privateData = JSON.stringify(data);
```
kalturaguy: it should be        KalturaBeacon.privateData = data
(we are not sending a string, but an object)

lib/Diagnostics/DiagnosticsAlertsAnalyzer.js
```diff
@@ -179,11 +174,14 @@ class EntryDiagnostics {
                     }
                 }
             });
+
+            if (this.alertsToSend.length) {
+                this.emit('reportBeacon', this.alertsToSend);
```
kalturaguy: maybe you should also have this.alertsToRemove..., so this will solve the other problem...

lib/Diagnostics/Diagnostics.js
```diff
@@ -40,16 +94,176 @@ class Diagnostics {
...
+            })
+    }
+
+    removeBeacon(key) {
+        let removedAlert = this.reportedAlerts.get(key);
+        let deleteResult = this.reportedAlerts.delete(key);
+        // Only if an alert was actually removed from map, update beacon
+        if (deleteResult) {
+            this._logger.debug(`[removeAlert] Alert: ${removedAlert.name}`);
+            return this._updateStreamHealth.call(this)
```
kalturaguy: in case of multiple removals, then you will call multiple time to API...

lib/Diagnostics/DiagnosticsAlertsAnalyzer.js
gadyaari: done

lib/Diagnostics/Diagnostics.js
gadyaari: done

lib/BackendClient.js
gadyaari: Tested. Server doesn't not support receiving just any object. has to be a json string.

PR #511, opened 2017-08-10
1.17 -temp-PR-PLAT-7830 

lilachmaliniak:
temp PR for Noam’s review and remove later.


lib/BackendClient.js
```diff
@@ -348,20 +350,40 @@ var BackendClient = (function(){
...
-        client.liveStream.update( function(result, err, headers) {
-            printAPIResponse(result, headers);
-            if(err || result.objectType === 'KalturaAPIException') {
-                logger.debug("[%s] Failed to update entry duration for [%s] with error [%s]", entryId,  ErrorUtils.error2string(err));
-                deferred.reject();
+    function updateEntryDuration(entryId, cumulativeDurationMs, recordedEntryId, entryServerNodeId) {
+        let deferred = Q.defer();
+        var mediaEntry = new kalturaVO.KalturaMediaEntry();
+        mediaEntry.msDuration = cumulativeDurationMs;
+        var serverEntryNode = new kalturaVO.KalturaLiveEntryServerNode();
```
noam-arad: for consistency I think the var name should be called "entryServerNode"

lib/entry/LiveEntry.js
```diff
@@ -130,6 +131,30 @@ function saveFlavorDownloadersConfiguration() {
     });
 }
 
+function getEntryServerNodeId() {
```
noam-arad: the method getLiveEntryServerNodes can also accept a filter of type KalturaEntryServerNodeFilter which you can put serverTypeEqual = this.entryServerType, then you know the entryServerNode you receive is the correct one and it will simplify this method.

lib/entry/StateManager.js
```diff
@@ -20,7 +20,7 @@ module.exports = function(liveEntry) {
     let lastServerUpdate = null;
     let lastLiveTimestamp = new Date();
     let stateMachineMessagePattern = "Event '%s' raised for entry: Changing state from '%s' to '%s'";
-    let lastStreamInfoUpdateTime = 0;
+    let lastStreamInfoUpdateTime = new Date(1970,1,1);
```
noam-arad: I think you meant new Date(0), this is more resistant to future changes to javascript implementation changes.

lib/entry/LiveEntry.js
```diff
@@ -130,6 +131,30 @@ function saveFlavorDownloadersConfiguration() {
...
+    var getLiveEntryServerNodeIdPromise =
```
noam-arad: Also this looks like a duplicate of StateManager::getEntryServerNodeId
I suggest to make the suggested change and move it to a shared code location

lib/kaltura-client-lib/KalturaVO.js
```diff
@@ -9960,10 +9973,12 @@ util.inherits(KalturaLiveEntryScheduleResource, KalturaScheduleResource);
 
 /**
  * @param streams array parameters of the stream we got.
+ * @param recordeInfo, array containing recording params (e.g. recordedEntryId, duration)
```
noam-arad: recordedInfo => recordingInfo

lib/recording/RecordingEntrySession.js
```diff
@@ -27,7 +27,7 @@ const recordingMaxDurationInMs = config.get('recording').recordingMaxDurationInH
 const unknownEntry = "UNKNOWN";
 
 class RecordingEntrySession {
-    constructor(entryObj, mp4FilesQueue = null) {
+    constructor(entryObj, entryServerNodeId, mp4FilesQueue = null) {
```
noam-arad: I saw that RestoreRecordingScript.js and Recording-spec.js have references to this constructor and don't seem to be updated. I don't know if those files are still relevant.

liveRecorder/Tasks/ConcatinationTask.py
```diff
@@ -124,6 +132,52 @@ def get_flavor_id(self, url_postfix, single_flavor):
...
+                        self.logger.debug('only one dc returned from call to server nodes list. Recording job will be processed')
+                        return True
+                    elif len(response_list.objects) == 0:
+                        if server_type.value == KalturaEntryServerNodeType.LIVE_PRIMARY:
+                            self.logger.debug('recording job was processed by LIVE BACKUP DC')
+                        else:
+                            self.logger.debug('recording job was processed by LIVE PRIMARY DC')
+                        return False
+                    else:
+                        for server_node in response_list.objects:
```
noam-arad: should be called entry_server_node

lib/BackendClient.js
lilachmaliniak: agree. done

lib/entry/LiveEntry.js
lilachmaliniak: updated

lib/entry/LiveEntry.js
lilachmaliniak: yes. As I moved the function from StateManager to LiveEntry as I think it is the correct place. Removed it from stateManager.

lib/entry/StateManager.js
lilachmaliniak: corrected.

lib/kaltura-client-lib/KalturaVO.js
lilachmaliniak: fixed

lib/BackendClient.js
```diff
@@ -348,20 +350,40 @@ var BackendClient = (function(){
...
+        serverEntryNode.entryId = entryId;
+        let recordingInfo = new kalturaVO.KalturaLiveEntryServerNodeRecordingInfo();
+        recordingInfo.recordedEntryId = recordedEntryId;
+        recordingInfo.duration = cumulativeDurationMs;
+        let recordingInfoArray = [];
+        recordingInfoArray.push(recordingInfo);
+        serverEntryNode.recordingInfo = recordingInfoArray;
+        let durationSec = (cumulativeDurationMs/1000).toFixed(0);
+        let durationString = ('0'+Math.floor(durationSec/3600) % 24).slice(-2)+':'+('0'+Math.floor(durationSec/60) % 60).slice(-2)+':'+('0' + durationSec % 60).slice(-2);
```
kalturaguy: this should go to utils

lib/BackendClient.js
```diff
@@ -237,29 +237,31 @@ var BackendClient = (function(){
         return deferred.promise;
     }
 
-    function registerEntryInDatabase(entryObject, state, event) {
+    function registerEntryInDatabase(entryObj, state, event) {
```
kalturaguy: why are you changing this method?  
1.this is not good, since we call this method every minute....
2. you should never call createRecordedEntry since it will cause race between the two DC

lib/BackendClient.js
```diff
@@ -348,20 +350,40 @@ var BackendClient = (function(){
...
-        var deferred = Q.defer();
-        KalturaMediaEntry.msDuration = cumulativeDurationMs;
-        client.liveStream.update( function(result, err, headers) {
-            printAPIResponse(result, headers);
-            if(err || result.objectType === 'KalturaAPIException') {
-                logger.debug("[%s] Failed to update entry duration for [%s] with error [%s]", entryId,  ErrorUtils.error2string(err));
-                deferred.reject();
+    function updateEntryDuration(entryId, cumulativeDurationMs, recordedEntryId, entryServerNodeId) {
+        let deferred = Q.defer();
+        var mediaEntry = new kalturaVO.KalturaMediaEntry();
```
kalturaguy: use let

lib/BackendClient.js
```diff
@@ -389,13 +411,15 @@ var BackendClient = (function(){
         client.liveStream.update(null, entryObj.entryId, KalturaMediaEntry);
         client.entryServerNode.listAction(null, EntryServerNodeFilter);
         client.liveStream.registerMediaServer(null, entryObj.entryId, mediaServerHostname, entryObj.serverType, applicationName, streamState);
+        client.liveStream.createRecordedEntry(null, entryObj.entryId, entryObj.serverType, streamState);
```
kalturaguy: why are you calling createRecordedEntry?  registerMediaServer should do it for you

lib/entry/LiveEntry.js
```diff
@@ -19,6 +19,7 @@ const Diagnostics = require('../Diagnostics/Diagnostics');
 const DiagnosticsAlertsAnalyzer = require('./../Diagnostics/DiagnosticsAlertsAnalyzer');
 const BackendClient = require('../BackendClientFactory.js').getBackendClient();
 const PromUtils = require('../utils/promise-utils');
+const backendClient = require('../BackendClientFactory.js').getBackendClient();
```
kalturaguy: duplicate code from two lines above.

lib/BackendClient.js
```diff
@@ -237,29 +237,31 @@ var BackendClient = (function(){
...
+                var errMsg=util.format(`[${entryObj.entryId}] Failed to register media from server: ${ErrorUtils.error2string(err)} ${results}`);
+                logger.error(errMsg);
+                deferred.reject(new Error(errMsg));
             }
-        }, entryObject.entryId, mediaServerHostname, entryObject.serverType, applicationName , state);
-
-        //Remove impersonation
-        client.setPartnerId(partnerId);
-        return deferred.promise;
+        });
```
kalturaguy: //Remove impersonation		
 -        client.setPartnerId(partnerId);

lib/recording/RecordingManager.js
```diff
@@ -345,10 +345,18 @@ class RecordingManager {
                                 logger.error(`[${entryId}-${recordingEntrySession.recordedEntryId}][stopRecording]: recorded entry is unknown, skip creating soft link`);
                                 return Q.resolve()
                             }
-                            if (recordingEntrySession.serverType !== kalturaTypes.KalturaEntryServerNodeType.LIVE_PRIMARY) {
-                                logger.info(`[${entryId}-${recordingEntrySession.recordedEntryId}][stopRecording] entry is backup, skip creating soft link`);
-                                return Q.resolve()
-                            }
+                        })
+                        .then(() => {
+                            let dcPath = path.join(recordingEntrySession.recordingSessionPath, 'dc');
```
kalturaguy: i suggest instead of saving "dc" file to save json file with all sorts of metadata (for now it's only dc)
but maybe in the future we will need more stuff)

lib/recording/RecordingEntrySession.js
lilachmaliniak: fixed not tested yet

liveRecorder/Tasks/ConcatinationTask.py
lilachmaliniak: renamed

lib/BackendClient.js
```diff
@@ -355,18 +356,18 @@ var BackendClient = (function(){
         var mediaEntry = new kalturaVO.KalturaMediaEntry();
         mediaEntry.msDuration = cumulativeDurationMs;
         var serverEntryNode = new kalturaVO.KalturaLiveEntryServerNode();
```
noam-arad: This should also be changed to entryServerNode

lib/entry/StateManager.js
```diff
@@ -131,7 +131,11 @@ module.exports = function(liveEntry) {
     }
 
     function updateStreamInfo(currentTime) {
-        if ((currentTime - lastStreamInfoUpdateTime) >= streamInfoUpdateInterval) {
+        let timeDiff = Math.abs(currentTime.getTime() - lastStreamInfoUpdateTime.getTime());
+
+        if (+timeDiff >= +streamInfoUpdateInterval) {
+            // todo: important to decide whether to remove or leave call to getEntryServerNodeId, since it is called directly from liveEntry
+            // can the server node id change during live session after short stream stop start (without stopping session)?
             return getEntryServerNodeId()
```
noam-arad: since you got rid of the function getEntryServerNodeId this is either dead code or a potential fatal.

lib/kaltura-client-lib/KalturaVO.js
noam-arad: still not fixed in the comments

lib/BackendClient.js
```diff
@@ -348,20 +351,40 @@ var BackendClient = (function(){
...
-        client.liveStream.update( function(result, err, headers) {
-            printAPIResponse(result, headers);
-            if(err || result.objectType === 'KalturaAPIException') {
-                logger.debug("[%s] Failed to update entry duration for [%s] with error [%s]", entryId,  ErrorUtils.error2string(err));
-                deferred.reject();
+    function updateEntryDuration(entryId, cumulativeDurationMs, recordedEntryId, entryServerNodeId) {
+        let deferred = Q.defer();
+        var mediaEntry = new kalturaVO.KalturaMediaEntry();
+        mediaEntry.msDuration = cumulativeDurationMs;
+        var serverEntryNode = new kalturaVO.KalturaLiveEntryServerNode();
```
noam-arad: this line should declare entryServerNode since that's what you call this var later on.

lib/kaltura-client-lib/KalturaVO.js
lilachmaliniak: done

lib/entry/LiveEntry.js
lilachmaliniak: removed

lib/BackendClient.js
lilachmaliniak: done

lib/BackendClient.js
lilachmaliniak: done

lib/entry/StateManager.js
lilachmaliniak: removed

lib/BackendClient.js
lilachmaliniak: separated the call to registerEntryInDatabase from craeteRecordedEntry (new method)

lib/BackendClient.js
lilachmaliniak: fixed

lib/BackendClient.js
lilachmaliniak: moved

lib/BackendClient.js
lilachmaliniak: moved

lib/BackendClient.js
lilachmaliniak: separated call to registerMediaServer from createRecordedEntry (new method)

lib/recording/RecordingManager.js
lilachmaliniak: removed replaced dc with metadata.json

lib/BackendClient.js
```diff
@@ -348,20 +349,40 @@ var BackendClient = (function(){
...
-            printAPIResponse(result, headers);
-            if(err || result.objectType === 'KalturaAPIException') {
-                logger.debug("[%s] Failed to update entry duration for [%s] with error [%s]", entryId,  ErrorUtils.error2string(err));
-                deferred.reject();
+    function updateEntryDuration(entryId, cumulativeDurationMs, recordedEntryId, entryServerNodeId) {
+        let deferred = Q.defer();
+        let mediaEntry = new kalturaVO.KalturaMediaEntry();
+        mediaEntry.msDuration = cumulativeDurationMs;
+        let entryServerNode = new kalturaVO.KalturaLiveEntryServerNode();
+        entryServerNode.entryId = entryId;
```
kalturaguy: why not pass the entryServerNode we are updating?  we already have it

lib/entry/LiveEntry.js
```diff
@@ -130,6 +131,23 @@ function saveFlavorDownloadersConfiguration() {
     });
 }
 
+function getEntryServerNodeId() {
+    let entryServerNodeFilter = new kalturaVO.KalturaEntryServerNodeFilter();
+    entryServerNodeFilter.serverTypeEqual = this.entryServerType;
+    var getLiveEntryServerNodeIdPromise =
```
kalturaguy: i think that the call to the API should be BackendClient.getEntryServerNodeByType(this.entryId,this.entryServerType)
(rename getSingleEntryServerNode to getEntryServerNodeByType and  move the KalturaEntryServerNodeFilter into inside the getEntryServerNodeByType function)

this is because the filter is internal implmentation of the backendclient

lib/BackendClient.js
```diff
@@ -433,6 +457,41 @@ var BackendClient = (function(){
         return deferred.promise;
     }
 
+    function createRecordedEntry(entryObj)  {
+        let deferred = Q.defer();
+        client.startMultiRequest();
+        // entryServerNode.listAction() parameter
+        let entryServerNodeFilter = new kalturaVO.KalturaEntryServerNodeFilter();
+        entryServerNodeFilter.entryIdEqual = entryObj.entryId;
+        let streamState = '{1:result:objects:0:status}';
```
kalturaguy: again... we already know the entryServerNodeId... no need to call list again?

lib/entry/StateManager.js
```diff
@@ -131,26 +126,21 @@ module.exports = function(liveEntry) {
...
+
+        if (+timeDiff >= +streamInfoUpdateInterval) {
+            if (liveEntry.isMultipleAudio) {
+                logger.debug("Stream is Multiple Audio. Skipping server update");
+                lastStreamInfoUpdateTime = new Date();
+                return Q.resolve();
+            }
+            else {
+                return updateFlavorsToServer(liveEntry.entryId, liveEntry.entryServerNodeId, liveEntry.flavorsDownloaders);
+            }
```
kalturaguy: you removed the   liveEntry.entryServerNodeId = null;		 in case of a catch.

this was added by Yossi on purpose... don't remember the bug that it fixed, please added it again and talk with him

lib/recording/RecordingEntrySession.js
```diff
@@ -47,7 +49,8 @@ class RecordingEntrySession {
         let now = new Date();
         let entryId = entryObj.entryId;
         let recordedEntryId = entryObj.recordedEntryId;
-        if (entryObj.recordStatus === kalturaTypes.KalturaRecordStatus.APPENDED){
+        this.entryServerNodeId = entryServerNodeId;
```
kalturaguy: i think there were situations where entryServerNodeId has changed during the stream, also entryObj already contains entryServerNodeId... so it's duplication

lib/recording/RecordingEntrySession.js
```diff
@@ -358,27 +375,58 @@ class RecordingEntrySession {
...
                 return Q.resolve();
             })
 
     }
 
-    changeRecordedEntry(recordedEntryId){
+    changeRecordedEntry(recordedEntryId) {
+        // Todo: decide to use only one member containg recordedEntryId be it the this.entryObject or the this.recordedEntryId
         this.recordedEntryId = recordedEntryId;
+        this.entryObject.recordedEntryId = recordedEntryId;
```
kalturaguy: sounds like duplication... we should keep and reference only one recordedEntryId value (in entryObject)

liveRecorder/Tasks/ConcatinationTask.py
```diff
@@ -124,6 +132,48 @@ def get_flavor_id(self, url_postfix, single_flavor):
 
         return flavor_id
 
+    def is_processing_required(self):
+        # read metadata file && get entry server nodes to decide whether to process the job or skip it
+        with open(self.metadata_full_path, "r") as dc_file:
+            data = dc_file.read()
+            if data:
+                metadata = json.loads(data)
```
kalturaguy: there should be a catch on the json.loads

lib/entry/StateManager.js
lilachmaliniak: moved to getEntryServerNodeId() LiveEntry.js

lib/BackendClient.js
```diff
@@ -255,7 +256,7 @@ var BackendClient = (function(){
                 logger.error(msg);
                 deferred.reject(new Error(msg));
             }
-        }, entryObject.entryId, mediaServerHostname, entryObject.serverType, applicationName , state);
+        }, entryObject.entryId, mediaServerHostname, entryObject.serverType, applicationName , streamState, false);
```
lilachmaliniak: added

lib/BackendClient.js
```diff
@@ -348,20 +349,40 @@ var BackendClient = (function(){
...
+            else {
+                let errMsg = util.format(`[${entryId}-${recordedEntryId}] Failed to update entry duration to [${durationString}]=[${cumulativeDurationMs} msec], entryServerNodeId:[${entryServerNodeId}] ${ErrorUtils.error2string(err)}. ${results}`);
+                logger.error(errMsg);
+                return deferred.reject(new Error(errMsg));
+            }
+        });
 
-        }, entryId, KalturaMediaEntry);
         return deferred.promise;
+
```
lilachmaliniak: deleted

lib/BackendClient.js
```diff
@@ -433,6 +457,41 @@ var BackendClient = (function(){
...
+        client.doMultiRequest((results, err, headers)=> {
+            printAPIResponse(results, headers);
+
+            if (iskMultiRequestSuccessful(results, err, 2)) {
+                let newRecordedEntryId = results[1].recordedEntryId;
+                logger.info(`[${entryObj.entryId}] Successfully retrieved new recordedEntryId:[${newRecordedEntryId}], replacing [${entryObj.recordedEntryId}]`);
+                deferred.resolve(newRecordedEntryId);
+            }
+            else {
+                let errMsg = util.format(`[${entryObj.entryId}] Failed to retrieve live stream entry info for from server: ${ErrorUtils.error2string(err)}. ${results}`);
```
lilachmaliniak: rephrased

lib/BackendClient.js
```diff
@@ -443,6 +502,20 @@ var BackendClient = (function(){
...
+    BackendClient.getSingleEntryServerNode = function(entryId, filter) {
+        return getLiveEntryServerNodes(entryId, filter)
+            .then((liveEntryServerNodes)=> {
+                if (liveEntryServerNodes.length > 0) {
+                    logger.info(`[${entryId}], received entry server node, id:[${liveEntryServerNodes[0].id}]`);
+                    return Q.resolve(liveEntryServerNodes[0]);
+                }
+                else {
+                    logger.error(`[${entryId}] failed to get entry server node`);
+                    Q.reject(`Couldn't find entryServerNode Id, of DC provisioning live entry:[${entryId}]`);
```
lilachmaliniak: rephrased

lib/entry/LiveEntry.js
```diff
@@ -130,6 +131,23 @@ function saveFlavorDownloadersConfiguration() {
     });
 }
 
+function getEntryServerNodeId() {
+    let entryServerNodeFilter = new kalturaVO.KalturaEntryServerNodeFilter();
```
lilachmaliniak: move filter to backendClient.js

lib/entry/LiveEntry.js
lilachmaliniak: yep.done

lib/entry/LiveEntry.js
```diff
@@ -130,6 +131,23 @@ function saveFlavorDownloadersConfiguration() {
...
+    entryServerNodeFilter.serverTypeEqual = this.entryServerType;
+    var getLiveEntryServerNodeIdPromise =
+           BackendClient.getSingleEntryServerNode(this.entryId, entryServerNodeFilter)
+            .then((entryServerNode)=> {
+                this.entryServerNodeId = entryServerNode.id;
+            })
+            .catch(function (err) {
+                // todo: check if set entryServerNodeId value to 0 here is correct???
+                this.entryServerNodeId = 0;
```
lilachmaliniak: yep .0 replaced by null

lib/entry/LiveEntry.js
```diff
@@ -173,37 +198,42 @@ LiveEntry.prototype.start = function() {
...
+            return stopLiveInRecordingPromise.then(()=> {
+                    return this.cleanup();
+                })
+                .finally(()=> {
+                    let stopPromise = this.FSM.stop();
                     return stopPromise
-                        .catch(function(err) {
-                            that.logger.error("Error calling FSM.stop: %s" ,ErrorUtils.error2string(err));
+                        .catch((err)=> {
+                            this.logger.error("Error calling FSM.stop: %s" ,ErrorUtils.error2string(err));
```
lilachmaliniak: ok.done

lib/recording/RecordingEntrySession.js
lilachmaliniak: removed this.recordedEntryId. Now using only this.entryObj.recodedEntryId.

lib/recording/RecordingEntrySession.js
lilachmaliniak: removed this.recordedEntryId

lib/recording/RecordingEntrySession.js
lilachmaliniak: no. entryObj doesn't include entryServerNodeId. Noam updated BE code so it won't happen.

lib/entry/StateManager.js
```diff
@@ -131,26 +126,21 @@ module.exports = function(liveEntry) {
...
-                    }
-                })
-                .catch(function (err) {
-                    logger.warn("entryServerNodeId not found, stream info will not update! Error: %s", ErrorUtils.error2string(err));
-                    liveEntry.entryServerNodeId = null;
-                    return Q.reject(err);
-                });
+        let timeDiff = Math.abs(currentTime.getTime() - lastStreamInfoUpdateTime.getTime());
+
+        if (+timeDiff >= +streamInfoUpdateInterval) {
```
lilachmaliniak: yep. removed

lib/recording/RecordingEntrySession.js
```diff
@@ -358,27 +388,58 @@ class RecordingEntrySession {
...
+        this.recordingSessionPath = this.getRecordingSessionPath(recordedEntryId);
         this.logger = loggerBase.getLogger('RecordingEntrySession', '[' + this.recordingSessionId + '] ');
         return this.playlistGenerator.setBasePath(this.recordingSessionPath);
     }
 
     getRecordingSessionPath(recordedEntryId) {
         return path.join(recordingRootPath, this.modeDirName, this.entryId.slice(-1), this.entryId, recordedEntryId);
     }
+
+    checkIfSplitSession(currentTime) {
```
lilachmaliniak: name changed to checkIfRecordedEntryIdChanged() which is actually what this method do.

lib/recording/RecordingEntrySession.js
```diff
@@ -358,27 +388,58 @@ class RecordingEntrySession {
...
+        let timeElapsed = Math.abs(currentTime.getTime - this.lastGetRecordedEntryIdTime.getTime());
+
+        if (+timeElapsed >= +getEntryInfoIntervalMSec) {
```
lilachmaliniak: as above. No harm done by '+', yet removed it.

lib/playlistGenerator/PlaylistGenerator.js
```diff
@@ -66,9 +65,10 @@ function PlaylistGenerator (entryObject, options) {
     that.persistPromise = qResolved;
 
 
-
     if (_.has(options, 'recording') && options.recording == true){
+        that.loggerInfo = `[${entryObject.entryId}-${entryObject.recordedEntryId}]`;
         that.playListFormat = 'recording';
+        that.recordedEntryId = entryObject.recordedEntryId;
```
kalturaguy: doesn't sound logic that the playlistgenerator should keep recordedEntryId

lib/entry/LiveEntry.js
```diff
@@ -144,8 +158,15 @@ LiveEntry.prototype.start = function() {
             return this.playlistGenerator.finalizeStart();
         })
         .then(()=> {
+            return getEntryServerNodeId.call(this)
```
kalturaguy: don't we already know the entryServerNodeId at this point?

lib/recording/RecordingManager.js
```diff
@@ -62,7 +62,7 @@ class RecordingManager {
         }
         if (!this.recordingList[entryId]) {
             logger.info(`[${entryObj.entryId}-${entryObj.recordedEntryId}] Create new recording session`);
-            let recordingEntrySession = new RecordingEntrySession(entryObj, mp4FilesQueue);
+            let recordingEntrySession = new RecordingEntrySession(entryObj, entryServerNodeId, mp4FilesQueue);
```
kalturaguy: it's not a good thing to pass the entryServerNodeId in the CTOR since entryServerNodeId could change during recorded session (stop/start)

tests/unit/Recording-spec.js
```diff
@@ -12,6 +12,8 @@ const sinon = require('sinon');
 const config = require('./../../common/Configuration');
 const kalturaTypes = require('../../lib/kaltura-client-lib/KalturaTypes');
 const qio = require('q-io/fs');
+const entryServerNodeId = '1111';
```
kalturaguy: ?????

liveRecorder/Tasks/UploadTask.py
```diff
@@ -93,6 +93,43 @@ def append_recording_handler(self, file_full_path, flavor_id, is_first_flavor):
...
+    def update_recording_status(self):
+        try:
+            with open(self.metadata_full_path, "r") as dc_file:
+                data = dc_file.read()
+                if data:
+                    metadata = json.loads(data)
+                    entry_server_node_id = -1
+                    try:
+                        entry_server_node_id = int(metadata['entryServerNodeId'])
+                        self.logger.debug(
```
kalturaguy: why do we need it if you are doing get get_server_entry_nodes_list anyways?
you will get only two ids in the response and the you know if the serverNode is on same DC...

liveRecorder/Tasks/ConcatinationTask.py
```diff
@@ -124,6 +133,100 @@ def get_flavor_id(self, url_postfix, single_flavor):
 
         return flavor_id
 
+    def is_processing_required(self):
+        # read metadata file && get entry server nodes to decide whether to process the job or skip it
+        should_process = True
+        local_dc_type_str = 'UNKNOWN'
+        remote_dc_type_str = 'UNKNOWN'
+        try:
+            with open(self.metadata_full_path, "r") as dc_file:
```
kalturaguy: this function is TOO long... also the logic can be simplified....  just do a generic try except...
also there is a log of code there that is useless like local_dc_type_str or remote_dc_type_str

liveRecorder/Tasks/ConcatinationTask.py
```diff
@@ -124,6 +133,100 @@ def get_flavor_id(self, url_postfix, single_flavor):
...
+                                for recorded_entry_info in entry_server_node.recordingInfo:
+                                    if entry_server_node.recordingInfo and recorded_entry_info.recordedEntryId == self.recorded_id:
+                                        recording_status = KalturaEntryServerNodeRecordingStatus(entry_server_node.recordingInfo.recordingStatus)
+                                        if recording_status == KalturaEntryServerNodeRecordingStatus.ON_GOING:
+                                            max_duration = recorded_entry_info.duration
+                                            id_of_processing_dc = entry_server_node.id
+                                            break
+                                        duration_diff_sec = (local_duration - recorded_entry_info.duration) / 1000
+                                        duration_is_equal = abs(duration_diff_sec) <= self.duration_diff_tolerance_sec
+                                        if not duration_is_equal and recorded_entry_info.duration > local_duration:
```
kalturaguy: where is the code that favors the primary instead of backup if +-same duration?

lib/entry/LiveEntry.js
lilachmaliniak: no

lilachmaliniak: Anyway moved code back to StateManager, though that is useless for recordingManager call to start recording

lib/entry/StateManager.js
lilachmaliniak: code moved back though it was done due to entryServerNode being taken from entryInfo object received before call to registerEntryInDatabase... so EntryServerNodeId is undefined (null)

lib/entry/LiveEntry.js
lilachmaliniak: Added a separate call to get EntryServerNodeId from RecordingEntrySession

lib/playlistGenerator/PlaylistGenerator.js
lilachmaliniak: These changes were done to enable better debugging and a way to separate log messages related to live stream from recording in context of playlistGenerator. removed the recordingEntryId as class member, left the loggerInfo

liveRecorder/Tasks/ConcatinationTask.py
lilachmaliniak: took outside all the business logic

liveRecorder/Tasks/ConcatinationTask.py
lilachmaliniak: check now and you'll find it easily, in should_process_as_redundant_dc(), line 240

liveRecorder/Tasks/UploadTask.py
lilachmaliniak: The motivation for the status is to avoid processing job if stream is live. such as in following use case:
network throttling caused recording session to end in DC#1 but stream is still live in DC#2. According to Yossi, if I upload the VOD from DC#1 while DC#2 is live it can cause problem. So I use it mainly to verify that I don't process and upload VOD while the stream is live and recording on same RecordedEntryId

tests/unit/Recording-spec.js
lilachmaliniak: this is part of unit tests!!!!

lib/BackendClient.js
```diff
@@ -437,6 +459,40 @@ var BackendClient = (function(){
         return deferred.promise;
     }
 
+    function createRecordedEntry(entryObj)  {
+        let deferred = Q.defer();
+        client.startMultiRequest();
+        // entryServerNode.listAction() parameter
+        let entryServerNodeFilter = new kalturaVO.KalturaEntryServerNodeFilter();
+        entryServerNodeFilter.entryIdEqual = entryObj.entryId;
```
kalturaguy: why are we filtering the entryServerNodeId?  don't we know that already?

lib/BackendClient.js
```diff
@@ -467,6 +523,36 @@ var BackendClient = (function(){
...
+
+        let entryServerNode = new kalturaVO.KalturaLiveEntryServerNode();
+        let recordingInfo = new kalturaVO.KalturaLiveEntryServerNodeRecordingInfo();
+        recordingInfo.recordedEntryId = recordedEntryId,
+        recordingInfo.recordingStatus = recordingStatus;
+        let recordingInfoArray = [];
+        recordingInfoArray.push(recordingInfo);
+        entryServerNode.recordingInfo = recordingInfoArray;
+        
+        client.entryServerNode.update((result, err, headers) => {
```
kalturaguy: do we need impersonation here?

lib/entry/StateManager.js
```diff
@@ -143,14 +140,27 @@ module.exports = function(liveEntry) {
                         return updateFlavorsToServer(liveEntry.entryId, liveEntry.entryServerNodeId, liveEntry.flavorsDownloaders);
                     }
                 })
```
kalturaguy: you remove thie catch code which is important in case if entryServerNodeId changed  getEntryServerNodeId() will not know to re-take the  new entryServerNodeId

lib/BackendClient.js
```diff
@@ -477,6 +563,23 @@ var BackendClient = (function(){
...
     };
+    
+    BackendClient.getSingleEntryServerNode = function(entryId, entryServerType) {
+        let filter = new kalturaVO.KalturaEntryServerNodeFilter();
+        filter.serverTypeEqual = entryServerType;
+        return getLiveEntryServerNodes(entryId, filter)
+            .then((liveEntryServerNodes)=> {
+                if (liveEntryServerNodes.length > 0) {
+                    logger.info(`[${entryId}], received entry server node, id:[${liveEntryServerNodes[0].id}]`);
+                    return Q.resolve(liveEntryServerNodes[0]);
```
kalturaguy: why are you taking the first one? your live controller can acaully be using the second one..

lib/entry/StateManager.js
```diff
@@ -143,14 +140,27 @@ module.exports = function(liveEntry) {
...
-                    return Q.reject(err);
-                });
         }
-        else
+        else {
             return Q.resolve();
+        }
+    }
+
+    function getEntryServerNodeId() {
```
kalturaguy: don't understand why are chanign this code..., we are adding new bugs here getSingleEntryServerNode returns the first entryServerNdoe instead of what the old code did...

lib/BackendClient.js
lilachmaliniak: Not related to entryServerNodeId

lib/BackendClient.js
lilachmaliniak: because after stop start the providing DC can change with RecordedEntryId. We need to filter EntryServerNodes having different RecordingEntryId's recordingInfo... we cannot filter in list request because BE doesn't support that

lib/BackendClient.js
lilachmaliniak: No. If you think it's wrong need request Noam to change permission.

lib/BackendClient.js
lilachmaliniak: If we already filtered based on server type (LIVE_PRIMARY/LIVE_BACKUP) how can it be a 2nd one, however to make sure added also isLive = true and StatusIn 1,2

lib/entry/StateManager.js
lilachmaliniak: my mistake. Fixed

lib/entry/StateManager.js
lilachmaliniak: The old code was not efficient and it is tested. the business logic to get the exact relevant EntryServerNode object is done in BackendClient.js

PR #512, opened 2017-08-14
use new client and add better logging to backend client

noam-arad:


liveRecorder/KalturaClient/Client.py
```diff
@@ -25,25 +25,25 @@
 #
 # @ignore
 # ===================================================================================================
-from Plugins.Core import *
-from Base import *
+from __future__ import absolute_import
```
kalturaguy: do we have __future__ (python 2.7.13) in production?

liveRecorder/KalturaClient/Client.py
```diff
@@ -25,25 +25,25 @@
...
-import gzip
 import time
 import os
 
-from poster.streaminghttp import register_openers
-from poster.encode import multipart_encode
-import urllib2
+import requests
+from requests_toolbelt.multipart.encoder import MultipartEncoder
+import six
```
kalturaguy: not a big fan of adding more dependencies (what is this six library...)
will not need update install scripts + jess....

Rather not go that way

noam-arad: This is what the official Kaltura Client uses
This way every time we need a client change we can just take the official one instead of adding our own ad-hoc code
also some of our existing code changed the client to return the HTTP headers which is not needed since you can just get them by using "client.responseHeaders".
Moreover this adds logging straight into the client and it's actions.

kalturaguy: ok, but we need to update the deployment scripts/document for this dependencies.

PR #513, opened 2017-08-27
Fix alerts beacon handling to send each alert once

gadyaari:
1) Make sure each alert is only reported once.

### DIFF
--- a/lib/Diagnostics/Diagnostics.js
+++ b/lib/Diagnostics/Diagnostics.js
@@ -108,10 +108,9 @@ class Diagnostics extends EventEmitter {
     updateBeacons(beaconsModificationObj) {
         let isReported = false;
         let isRemoved = false;
-
         if (beaconsModificationObj.report) {
-            isReported = true;
-            this._reportBeacon(beaconsModificationObj.report);
+            // Only if new alerts were inserted to the map, update beacon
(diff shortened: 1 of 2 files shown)


lib/Diagnostics/Diagnostics.js
```diff
@@ -160,30 +165,34 @@ class Diagnostics extends EventEmitter {
...
         return Q.fcall(()=> {
             let streamHealth = StreamHealth.Good;
-            if (this.reportedAlerts.length) {
+            if (this.reportedAlerts.size) {
                 let isPoor = false;
                 let isFair = false;
-                _.forEach(this.reportedAlerts, a => {
-                    if (a.health === StreamHealth.Poor) {
+                this.reportedAlerts.forEach((value) => {
+                    if (value.health === StreamHealth.Poor) {
```
kalturaguy: why not streamHealth=Math.max(streamHealth,alert.health)?

gadyaari: because streamHealth is not an enum, but of type:
`const StreamHealth = {
    Good: 'Good',
    Fair: 'Fair',
    Poor: 'Poor'
};`

gadyaari: If I change it to numeric enum the dashboard would have to parse it to string.
I thought we wanted to save this necessity from it

PR #520, opened 2017-09-06
Fix entryServerNode type issues and erase ttl from beacon action

gadyaari:


lib/BackendClient.js
```diff
@@ -182,8 +178,16 @@ var BackendClient = (function(){
     }
 
     function getServerType(serverNodeId, resultReceived) {
-        return (resultReceived.primaryServerNodeId === serverNodeId) ? kalturaTypes.KalturaEntryServerNodeType.LIVE_PRIMARY :
-            kalturaTypes.KalturaEntryServerNodeType.LIVE_BACKUP;
+        // return (resultReceived.primaryServerNodeId === serverNodeId) ? kalturaTypes.KalturaEntryServerNodeType.LIVE_PRIMARY :
+        //     kalturaTypes.KalturaEntryServerNodeType.LIVE_BACKUP;
+        if (resultReceived.primaryServerNodeId === serverNodeId) {
+            return (serverNodeId === 0) ? kalturaTypes.KalturaEntryServerNodeType.LIVE_PRIMARY :
```
kalturaguy: this looks like a bug...  if serverNodeId is 0 it means we don't know it yet...

PR #529, opened 2017-09-27
merge to 1.19 fix in getSorterFileList

lilachmaliniak:
now handles error such as: recorded content dir not found


kalturaguy:
Did you test it?

PR #533, opened 2017-10-02
Refactor the beacon reported alerts sentfrom flavorDownloaders

gadyaari:


common/config/config.json.template
```diff
@@ -5,6 +5,7 @@
   "rootFolderPath" : "~/kLive_content/live",
   "oldContentFolderPath" : "~/kLive_content/archive",
   "requestTimeout" : 20000,
+  "flavorSourceId" : "32",
```
kalturaguy: not good, never estimate that

gadyaari: there's not way for me to know which flavor is ingested (source or not) without making another API call
and I don't wanna do it

lib/Diagnostics/DiagnosticsAnalyzer.js
```diff
@@ -178,28 +182,66 @@ class EntryDiagnostics extends EventEmitter{
...
+                        this._checkIncomingBitrate(input.IOPerformance.bitrate, this._getConfiguredDatarate(sourceStreamInfo.inputs[index]), indexNum);
                     }
                     if (input.Encoder.syncPTSData) {
-                        this._checkPtsSynchronization(diagObj.currentTime, input.Encoder.syncPTSData, indexNum);
+                        this._checkPtsSynchronization(sourceStreamInfo.currentTime, input.Encoder.syncPTSData, indexNum);
                     }
                 }
             });
 
+            if (diagObj.flavors && diagObj.flavors.length) {
```
lilachmaliniak: rename diagObj to sourceStreamInfo

PR #537, opened 2017-10-11
PLAT-7955: Add alert if streaming only to backup server

gadyaari:


lib/Diagnostics/Diagnostics.js
```diff
@@ -224,8 +225,8 @@ class Diagnostics extends EventEmitter {
         // Check should be performed only when diagnostics object is updated
         if (this.lastDiagnosticsObjTime === metaData.source.currentTime)
             return;
-
-        this._diagnosticsAnalyzer.runFullDiagnosticsAnalysis(metaData);
+        // beaconPrefix is the string casting of the server type
```
kalturaguy: i think it's better to pass entryObject so you have this info instead of passing multiple args

lib/entry/StateManager.js
```diff
@@ -162,27 +162,26 @@ module.exports = function(liveEntry) {
...
-                    liveEntry.entryServerNodeId = 0;
-                    return Q.reject(err);
-                });
-
-        return getLiveEntryServerNodeIdPromise;
+                }
+                return liveEntry.entryServerNodeId ? Q.resolve() : Q.reject("Couldn't find entryServerNodeId");
+            })
+            .catch(function (err) {
+                liveEntry.entryServerNodeId = 0;
```
kalturaguy: maybe remove entryServerNodeId = 0 ?

lib/entry/StateManager.js
```diff
@@ -162,27 +162,26 @@ module.exports = function(liveEntry) {
...
     }
 
-    function getEntryServerNodeId() {
-        var getLiveEntryServerNodeIdPromise =
-            liveEntry.entryServerNodeId ? Q.resolve()
-                : backendClient.getLiveEntryServerNodes(liveEntry.entryId)
-                .then(function (liveEntryServerNodes) {
-                    _.each(liveEntryServerNodes, function (liveEntryServerNode) {
+    function updateEntryServerNodesInfo() {
+        return backendClient.getLiveEntryServerNodes(liveEntry.entryId)
```
kalturaguy: you need to ask Noam/Yossi how optimize is this call

lib/Diagnostics/Diagnostics.js
gadyaari: done

PR #545, opened 2017-11-06
merge script to restore recording from backup

lilachmaliniak:
added new script to restore recording from backup DC

### DIFF
--- a/lib/recording/RestoreRecordingFromBackup.js
+++ b/lib/recording/RestoreRecordingFromBackup.js
@@ -22,12 +22,11 @@ const completedRecordingFolderPath = config.get('recording').completedRecordingF
 options
     .version('0.0.1')
     .option('-p, --path <path>', 'specify path')
-    .option('-d --duration <duration msec>', 'recording duration in msec')
     .parse(process.argv);
 
-const valid = ('path' in options) && ('duration' in options);
+const valid = ('path' in options);
 if (valid !== true) {
-    logger.error("Missing mandatory argument, path (recording content full path) and duration (recording duration, msec. Can be taken from stamp file) must be passed as args, in command line");
+    logger.error("Missing mandatory argument, path (recording content full path) must be passed as arg, in command line");
     process.exit(1);
 }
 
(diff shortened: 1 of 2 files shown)


lib/recording/RestoreRecordingFromBackup.js
```diff
@@ -0,0 +1,113 @@
...
+const mediaServerHostname = config.get('mediaServer').hostname;
+const persistenceFormat = require('../../common/PersistenceFormat');
+const kalturaTypes = require('../kaltura-client-lib/KalturaTypes');
+const completedRecordingFolderPath = config.get('recording').completedRecordingFolderPath;
+const unknownEntry = 'UNKNOWN';
+
+options
+    .version('0.0.1')
+    .option('-p, --path <path>', 'specify path')
+    .option('-d --duration <duration msec>', 'recording duration in msec')
```
kalturaguy: why do we need to pass duation?

lib/recording/RestoreRecordingFromBackup.js
```diff
@@ -0,0 +1,113 @@
...
+
+backendClient.getEntryInfo(entryId)
+    .then((entryObject)=> {
+        partnerId = entryObject.partnerId;
+        if (entryObject.recordedEntryId) {
+            recordedEntryId = entryObject.recordedEntryId;
+            return entryObject;
+        }
+        else {
+            backendClient.registerEntryInDatabase({entryId: entryId, serverType: mediaServerIndex, partnerId: partnerId}, liveEntryStatus, 'play')
```
kalturaguy: you forgot "return" to avoid race

liveRecorder/Config/config.ini
```diff
@@ -1,35 +1,36 @@
 [default_config]
-recording_base_dir = /web/content/kLive/liveRecorder
+recording_base_dir = /Users/lilach.maliniak/kLive_content/liveRecorder/
```
kalturaguy: ????????

lib/recording/RestoreRecordingFromBackup.js
```diff
@@ -0,0 +1,113 @@
...
+const events = require("events");
+const qio = require('q-io/fs');
+const path = require('path');
+const _ = require('underscore');
+const glob = require('glob');
+const config = require('../../common/Configuration')
+//config.set('logToConsole', true);
+const ErrorUtils = require('../utils/error-utils');
+const PersistenceFormat = require('../../common/PersistenceFormat.js');
+const logger = require('../../common/logger').getLogger("RestoreRecordingFromBackup");
```
kalturaguy: the usage of the logger is not good here, since we wnt to display to console + we don't want to write into the logs of the actual live controller

lib/recording/RestoreRecordingFromBackup.js
lilachmaliniak: no need. replaced by reading stamp file

PR #551, opened 2017-11-09
Revert "Add CPU profiling support"

DBezemer:
This reverts commit e823ab5d19ac2eaa8f75dfec87bb7cafb393075d.
v8 profiler requires too many external dependencies that cannot be packaged for remote systems (as they are release and distribution specific).

As this code is only used on SIGUSR2 the proposal is to remove it altogether as you can attach a nodeJS profiler in DEV that performs the same task without needing code.


kalturaguy:
@DBezemer   we don't use this anymore, there are better tools :)

jessp01:
@kalturaguy - please create a new release so I can package it.

PR #552, opened 2017-11-12
merge PLAT-8238 to 1.22

lilachmaliniak:


liveRecorder/Tasks/ConcatinationTask.py
```diff
@@ -5,6 +5,7 @@
 import re
```
kalturaguy: why this task only? why not ALL a logic that belong on the BaseTask?

lilachmaliniak: The first task that get new job is ConcatenationTask. The retries, can be global for job handling cycle or per task in job handling cycle. We didn't discuss it so I chose to relate retires to whole job cycle. Let's discuss it.

kalturaguy: yes.. but you can have the same problem in the upload task?

lilachmaliniak: Anyway I moved logic to TaskBase. In the first place the retries counter was used for ALL jobs created on the entry. 
I suggest to decide which errors should have retries and which not, not just whether every task incoming should reset the retries. This will reduce useless reprocessing, in case of unrecoverable errors.

liveRecorder/Tasks/TaskBase.py
```diff
@@ -35,11 +35,23 @@ def __init__(self, param, logger_info):
...
+
+    def reset_retry_count(self, src):
+        try:
+            retries_file_path = os.path.join(src, 'retries')
+            if os.path.exists(retries_file_path):
+                with open(retries_file_path, "w") as retries_file:
+                    retries_file.write(self.failed_tasks_max_retries)
+
+        except Exception as e:
+            self.logger.error("Failed to reset retries count for %s: %s \n %s", src, str(e), traceback.format_exc())
```
kalturaguy: remove traceback and use exc_info=True

liveRecorder/Tasks/ConcatinationTask.py
```diff
@@ -126,6 +126,7 @@ def get_flavor_id(self, url_postfix, single_flavor):
 
     def run(self):
 
+        self.reset_retry_count(self.recording_path)
```
kalturaguy: both ConcatinationTask.py and UploadTask.py uses it, please move it to the base class

lilachmaliniak: moved reset_retry_count to BaseTask.py

PR #553, opened 2017-11-13
merge updated nginx version  to 1.22

lilachmaliniak:
1.12.2


packager/bin/build_nginx.sh
```diff
@@ -10,7 +10,7 @@ echo "${devRootDir}"
 
 ffmpegLibsDir=${ffmpegLibsDir:-${devRootDir}/liveDVR/node_addons/FormatConverter/build/FFmpeg}
 packagerDir="${devRootDir}/nginx-vod-module"
-nginxVersion=${nginxVersion:-1.20.0}
+nginxVersion=${nginxVersion:-1.12.2}
```
kalturaguy: don't think we support 1.12.2.

please consult with EranK which version we support

note that you have two versions:
nginx-vod-module
and the nginx.org

PR #557, opened 2017-11-21
merge liveRecorder script update to 1.23

lilachmaliniak:
after testing


lilachmaliniak:
updated liveRecorder script to always start using full path, instead of symlink.

PR #565, opened 2017-12-04
1.21 plat 7977 1 plat 8313

noam-arad:


packager/config/nginx.conf.live.conf.template
```diff
@@ -11,12 +11,19 @@ vod_dash_profiles urn:mpeg:dash:profile:isoff-live:2011;
 expires off;
 vod_upstream_location "";	
 
-location ~ \/(?<live_type>live|recording|live-stg)\/[^\/]+\/p\/(?<partnerId>\d+)\/e\/(?<entryId>[^\/]+(?<entryHash>\w))\/(sd\/(?<segmentDuration>\d+)\/)?t\/(?<token>[^\/]+) {
+map $user_type $overrideClipTo{
+	"user" "";
+	default '{"clipFrom":0, "clipTo":0}';
+}
+
+location ~ \/(?<live_type>live|recording|live-stg)\/[^\/]+\/p\/(?<partnerId>\d+)\/e\/(?<entryId>[^\/]+(?<entryHash>\w))\/(sd\/(?<segmentDuration>\d+)\/)?t\/(?<token>[^\/]+)(/type/(?<user_type>\w+))?\/ {
```
kalturaguy: the type should be before the token so we will sign it too

packager/config/nginx.conf.live.conf.template
```diff
@@ -11,12 +11,19 @@ vod_dash_profiles urn:mpeg:dash:profile:isoff-live:2011;
 expires off;
 vod_upstream_location "";	
 
-location ~ \/(?<live_type>live|recording|live-stg)\/[^\/]+\/p\/(?<partnerId>\d+)\/e\/(?<entryId>[^\/]+(?<entryHash>\w))\/(sd\/(?<segmentDuration>\d+)\/)?t\/(?<token>[^\/]+) {
+map $user_type $overrideClipTo{
+	"user" "";
+	default '{"clipFrom":0, "clipTo":0}';
```
kalturaguy: shouldn't it be clipTo: MaxInt?

packager/config/nginx.conf.live.conf.template
```diff
@@ -11,12 +11,19 @@ vod_dash_profiles urn:mpeg:dash:profile:isoff-live:2011;
...
+}
+
+location ~ \/(?<live_type>live|recording|live-stg)\/[^\/]+\/p\/(?<partnerId>\d+)\/e\/(?<entryId>[^\/]+(?<entryHash>\w))\/(sd\/(?<segmentDuration>\d+)\/)?t\/(?<token>[^\/]+)(/type/(?<user_type>\w+))?\/ {
 	#this is for backward compatability to fetch from wowza  
 	location ~ ^/m/(?<mediaServer>[^/]+)/live/legacy/p/\d+/e/[^/]+(/sd/[^/]+)?/t/[^/]+/(?<path>.*) {
 		proxy_pass http://${mediaServer}:1935/kLive/ngrp:${entryId}_1_all/$path$is_args$args;
 	}
 
+	vod_media_set_override_json $overrideClipTo;
```
kalturaguy: should be only in the live location (so it will not be applied to the recording too)

lib/entry/StateManager.js
```diff
@@ -121,8 +121,9 @@ module.exports = function(liveEntry) {
                                 pushManager.setObjectEnabled(liveEntry.entryId, false);
                                 return backendClient.registerEntryInDatabase(liveEntry.entryObject, KalturaLiveStatus.PLAYABLE, event)
                                     .then((newEntryObj) => {
+                                        const firstTime = (lastServerUpdate == null);
                                         lastServerUpdate = new Date();
-                                        liveEntry.entryObjectUpdated(newEntryObj);
+                                        liveEntry.entryObjectUpdated(newEntryObj, firstTime);
```
kalturaguy: what about scenario of two start/stop preview at the same sessioN?  firstTime will still be false

lib/entry/LiveEntry.js
```diff
@@ -318,6 +327,16 @@ LiveEntry.prototype._startDVR = function(entry)
     this.playlistGenerator.setDVRDuration(newPlayWindow);
 }
 
+LiveEntry.prototype._setClipTo = function(entry)
+{
+    this.playlistGenerator.setClipTo(new Date().getTime());
```
kalturaguy: this is not the best thing....

packager/config/nginx.conf.live.conf.template
noam-arad: After a short conversation with Eran he said he wants to change it so that if "clipTo" value is 0 it will just disregard.

PR #568, opened 2017-12-21
PLAT-8363: in order not to lose frames when recording is started via …

noam-arad:
…explicit live we keep the latest mp4 files and when recording is started we add the relevant ones to the recording and continue with regular recording

### DIFF
--- a/common/config/config.json.template
+++ b/common/config/config.json.template
@@ -117,5 +117,6 @@
        "recordingUpdateEntryTimeIntervalInMin" : 15,
        "recordingMaxAllowedMp4FilesQueueSize" : 100
      },
-   "explicitLivePushNotificationTemplateName": "EXPLICIT_LIVE_PUSH_NOTIFICATIONS"
+   "explicitLivePushNotificationTemplateName": "EXPLICIT_LIVE_PUSH_NOTIFICATIONS",
+   "recordingNotStartedMaxQueueSize": "10"
 }
--- a/common/config/config.json.template
+++ b/common/config/config.json.template
@@ -118,5 +118,5 @@
        "recordingMaxAllowedMp4FilesQueueSize" : 100
      },
    "explicitLivePushNotificationTemplateName": "EXPLICIT_LIVE_PUSH_NOTIFICATIONS",
-   "recordingNotStartedMaxQueueSize": "10"
+   "recordingNotStartedMaxQueueSize": "11"
 }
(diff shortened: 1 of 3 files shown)


common/config/config.json.template
```diff
@@ -117,5 +117,6 @@
        "recordingUpdateEntryTimeIntervalInMin" : 15,
        "recordingMaxAllowedMp4FilesQueueSize" : 100
      },
-   "explicitLivePushNotificationTemplateName": "EXPLICIT_LIVE_PUSH_NOTIFICATIONS"
+   "explicitLivePushNotificationTemplateName": "EXPLICIT_LIVE_PUSH_NOTIFICATIONS",
+   "recordingNotStartedMaxQueueSize": "10"
```
kalturaguy: if chunk duration is 6 then it's only 60 seconds, maybe put there 11?

lib/entry/FlavorDownloader.js
```diff
@@ -347,17 +349,29 @@ var FlavorDownloader = (function() {
...
                                 that.emit('newStreamFiles', that, duration / 1000);
-                                if (that.recordingEnable && that.entryObj.recordingStatus ==  kalturaTypes.KalturaRecordingStatus.ACTIVE) { //if recording is enable, and flavor is source, send event
-                                    try {
-                                        RecordingManager.addNewChunks(results.newMp4Files, that.entryId, that.flavor)
-                                            .catch((err)=> {
-                                                that.logger.error("Error while trying to run RecordingManager.addNewChunks %s", ErrorUtils.error2string(err));
-                                            });
-                                    } catch (err) {
-                                        that.logger.error("Error while trying to run RecordingManager.addNewChunks %s", ErrorUtils.error2string(err))
+                                if (that.recordingEnable)//if recording is enable, and flavor is source, send event
```
kalturaguy: i think this should be on a differnt function?

lib/entry/FlavorDownloader.js
```diff
@@ -509,6 +523,32 @@ var FlavorDownloader = (function() {
 
     }
 
+    function useSavedChunks(){
+        if (this.recordingStartTime)
+        {
+            for (let mp4Files of this.savedMp4Files)
+            {
+                let mp4FilesToUse = [];
+                for (let currMp4File of mp4Files)
```
kalturaguy: _.filter(mp4Files, currMp4file=> currMp4File.startTime + currMp4File.video.duration > this.recordingStartTime)

PR #578, opened 2018-02-05
PLAT-8554: update is_live_user only after clipTo is removed from play…

noam-arad:
…list
related to https://github.com/kaltura/server/pull/6753


lib/entry/LiveEntry.js
```diff
@@ -299,7 +299,8 @@ LiveEntry.prototype.entryObjectUpdated = function(entry)
         && origViewMode != kalturaTypes.KalturaViewMode.ALLOW_ALL)
     {
         this._startDVR(entry);
-        this._removeClipTo();
+        const persistePlaylistPromise = this._removeClipTo();
```
kalturaguy: better if you do this._removeClipTo().then ( ()=> this._updateIsLiveUser(entry, true) )

PR #582, opened 2018-03-06
changes to support more than just live server nodes

noam-arad:


lib/BackendClient.js
```diff
@@ -220,6 +220,7 @@ var BackendClient = (function(){
         client.startMultiRequest();
         var entryServerNodeFilter = new kalturaVO.KalturaEntryServerNodeFilter();
         entryServerNodeFilter.entryIdEqual = entryId;
+        entryServerNodeFilter.serverTypeIn = kalturaTypes.KalturaEntryServerNodeType.LIVE_BACKUP + ',' + kalturaTypes.KalturaEntryServerNodeType.LIVE_PRIMARY;
```
kalturaguy: why not instead do 
         var entryServerNodeFilter = new kalturaVO.KalturaLiveEntryServerNodeFilter();

PR #590, opened 2018-05-15
set file permission on newly created file

noam-arad:


david-winder-kaltura (approved):

liveRecorder/Tasks/ConcatinationTask.py
```diff
@@ -147,6 +148,8 @@ def run(self):
             self.download_chunks_and_concat(chunks, output_full_path)
             self.logger.info("Successfully concat %d files into %s", len(chunks), output_full_path)
         self.convert_ts_to_mp4(command)
+        if os.path.isfile(output_full_path):
+            os.chmod(output_full_path, NEW_FILE_PERMISSION)
```
kalturaguy: 1. why now? what is the bug?
2. why outside the for loop?

kalturaguy (changes requested):

PR #591, opened 2018-05-15
set file permission on newly created file

noam-arad:


liveRecorder/ts_to_mp4_convertor/main.c
```diff
@@ -566,6 +566,12 @@ int main(int argc, char **argv)
     {
         dispose(&conversion[i]);
     }
+
+    for (i=0;i<total_conversions;i++)
+    {
+        char* out_filename = argv[i*3+2];
+        chmod(out_filename, 0666);
+    }
```
kalturaguy: better if you move the chmod into
https://github.com/kaltura/liveDVR/blob/1.28/liveRecorder/ts_to_mp4_convertor/main.c#L311

noam-arad: inside the "FileConversion" struct we don't know the out file path and I couldn't discern if we have access to the actual FD

kalturaguy: then add it :)

kalturaguy: also i don't think we should change this... we need to figure out first why this work well in production and in QA not.  could be resolved by not adding code

noam-arad: generally speaking not setting a file permission after creating a file is a bad habit.
In this specific instance I would venture that the issue is caused in QA and as far as I know in production and will be exposed, because of our backend change, in the next deployment.
If it does work on env A and not env B probably has to do with permissions of the user that runs the service that runs the process but it is very hard to say, because it is hard to say you should always set file permissions explicitly

noam-arad: checked, it is not possible to get FD from AVIOContext.
I'll add the chmod inside the method.

PR #593, opened 2018-05-22
1.28 register to push race

noam-arad:


lib/PushManager.js
```diff
@@ -99,14 +94,23 @@ class PushManager {
...
     }
 
     _messageReceived(queueKey, msg)
     {
         logger.debug(`message received [${util.inspect(msg)}]`);
         if (this.registeredQueues[queueKey])
         {
-            const objectId = this.registeredQueues[queueKey].objectId;
+            let queueKeyHash = this.registeredQueues[queueKey];
+            const objectId = this.registeredQueues[queueKeyHash].objectId;
```
kalturaguy: why sometimes registeredQueues saves hash and sometimes an object... it's confusing...
why not always use queueKey as the key?

PR #594, opened 2018-05-22
disable push notifications for explicit live until race condition wit…

noam-arad:
…h on.connected is resolved


david-winder-kaltura (approved):
As temporary workaround

PR #604, opened 2018-08-22
fix restore from backup code

noam-arad:


lib/recording/RestoreRecordingFromBackup.js
```diff
@@ -71,14 +71,15 @@ qio.read(stampPath, 'r')
...
-                    return Q.delay(10000).backendClient.unregisterEntryInDatabase(reply)
-                        .then(()=> {
-                            logger.debug(`[${entryId}-${recordedEntryId}] after unregister media server`)
-                        }).catch( ()=> {
-                            //we don't actually care on unregister...
-                            return Q.resolve();
-                        });
-                });
+                    return Q.delay(10000).then(()=>{
+                        backendClient.unregisterEntryInDatabase(reply)
```
kalturaguy: you are missing return

PR #613, opened 2019-06-02
SUP-18414 dvr clip from

david-winder-kaltura:


lib/entry/LiveEntry.js
```diff
@@ -302,6 +302,7 @@ LiveEntry.prototype.entryObjectUpdated = function(entry)
         && origViewMode != kalturaTypes.KalturaViewMode.ALLOW_ALL)
     {
         this._startDVR(entry);
+        this._setClipFrom(entry.updatedAt * 1000, false);
```
noam-arad: Have a look at _removeClipTo, it forcefully persists the playlist now, that's why it returns a promise.
I think in this situation you should do the same

kalturaguy: this._setClipFrom((entry.updatedAt-config.get('minDurationForPlaybackInSec')) * 1000 , false);

PR #629, opened 2020-06-22
PSVAMB-14598: always return impersonated partnerId to media-server pa…

noam-arad:
…rtnerId


kalturaguy (approved):

lib/BackendClient.js
```diff
@@ -296,7 +295,7 @@ var BackendClient = (function(){
 
             deferred.resolve(res);
         }, filter);
-
+        client.setPartnerId(partnerId);
```
david-winder-kaltura, line 299: I think better to add comment - for the first glance it is not clear the partnerId is the -5 
As in https://github.com/kaltura/liveDVR/blob/c21a05c07efcc7757c8e57592235d6f3ab4d498c/lib/BackendClient.js#L476

david-winder-kaltura (approved):

